PR #6, opened 2020-09-15
Dev carles

### LINKED ISSUE #1: Simulation environment classes
I was thinking in trying to have an interface very similar to the ones of GYM environments (without actions) so later its easier to do some tests with them.

I think we should have an "environments" folder and have all the environments in there

My idea was to have something like:

```python
from environments.pendulum import Pendulum
env = Pendulum(mass=1, length=1)  # Instantiate class with its parameters
env.set(p=30, q=10)  # Fix some initial conditions. In this case p is 1-dim angle and q 1-dim ang vel
image, p, q = env.step(delta_time=0.1)  # Do a step of 0.1 seconds on the environment and get its info (environment is now in this new state)
```
What do you think @fedetask @charlio23?

### LINKED ISSUE #3: Implement pendulum simulation for dataset generation (black and white frames)
Implement a python class that is able to generate the time evolution of a 2-dimensional pendulum oscilatory movement.

It should obey the usual hamiltonian time evolution formulation. From the class instance one should be able to configure the properties of the pendulum (mass, length of string, initial conditions, etc) and output black and white frames.

charlio23:
Implemented environment abstract base class and pendulum system


environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+        Raises:
+            NotImplementedError: Class instantiation has no implementation
+        """
+        raise NotImplementedError
+
+    def step(self, dt=0.01):
+        """Performs a step in the environment simulator
+
+        Args:
+            dt (float, optional): [description]. Defaults to 0.01.
```
OleguerCanal: Put the description of the argument in [description]

environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        raise NotImplementedError
+
+    def draw(self):
```
OleguerCanal: No  abstract method decorator?

environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+        """Returns Caption of the environment state
+        """
+        raise NotImplementedError
+
+    def generateData(self, total_seconds, fps):
```
OleguerCanal: generate_data
Try to follow PEP8 convention of method naming: lowercase with words separated by underscores as necessary to improve readability.

environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+        """Generates dataset for current environemnt
+
+        Args:
+            total_seconds (int): Total duration of video (in seconds)
+            fps (int): Frame rate of generated data (frames per second)
+        
+        Returns dict containing input data and corresponding ground truth phase states
```
OleguerCanal: Return:
     (dict): bla bla

environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+
+        time_evol = 1./fps
+        total_images = total_seconds*fps
+        dt = 0.01
```
OleguerCanal: This should be an argument (maybe same as 1/fps)?

environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+        self.length = length
+        self.set(p, q)
+
+    def set(self, p, q):
+        self.p = p
+        self.q = q
+
+    def step(self, dt=0.01):
+
+        self.q += dt*(self.p/(self.mass*self.length*self.length))
```
OleguerCanal: Check if p or q are None

environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+        theta'' = -(g/l)*sin(theta)
+
+    """
+
+    def __init__(self, mass, length, p=None, q=None):
+        self.mass = mass
+        self.length = length
+        self.set(p, q)
+
+    def set(self, p, q):
```
OleguerCanal: Docstring missing

environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+        Args:
+            p (float): generalized momentum
+            q (float): generalized position
+
+        Raises:
+            NotImplementedError: Class instantiation has no implementation
+        """
+        raise NotImplementedError
+
+    def step(self, dt=0.01):
```
OleguerCanal: No abstractmethod decorator?

environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+    def set(self, p, q):
+        self.p = p
+        self.q = q
+
+    def step(self, dt=0.01):
+
+        self.q += dt*(self.p/(self.mass*self.length*self.length))
+        self.p += dt*-9.81*self.mass*self.length*np.sin(self.q)
+
+    def draw(self):
```
OleguerCanal: Missing docstring

environments/environments.py
```diff
@@ -0,0 +1,94 @@
+from abc import ABC, abstractmethod
+from PIL import Image, ImageDraw
+import numpy as np
+
+
+class Environment(ABC):
+
+    @abstractmethod
+    def set(self, p, q):
+        """Sets initial conditions for physical system
```
OleguerCanal, line 17: Docstring should be in imperative

environments/environments.py
```diff
@@ -0,0 +1,94 @@
...
+    Equations of movement are:
+
+        theta'' = -(g/l)*sin(theta)
+
+    """
+
+    def __init__(self, mass, length, p=None, q=None):
+        self.mass = mass
+        self.length = length
+        self.set(p, q)
```
OleguerCanal: Maybe this should be done in the abstract class since we'll have to copy this line in all the others otherwise (not very important though)

OleguerCanal (changes requested):
Can you move the Environment abstract class into a script called environment.py and the Pendulum class into a script named pendulum.py? This way we have each environment in a different script (cleaner)

environments/pendulum.py
```diff
@@ -0,0 +1,70 @@
...
+        self.q = q
+
+    def step(self, dt=0.01):
+        """Performs a step in the pendulum system
+
+        Args:
+            dt (float, optional): Time step run for the integration. Defaults to 0.01.
+
+        Raises:
+            TypeError: If p or q are None
```
OleguerCanal: AssertError right? Maybe TypeError also works, not sure

environments/pendulum.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        self.mass = mass
+        self.length = length
+        self.set(p, q)
+
+    def set(self, p, q):
+        """Sets initial conditions for pendulum
+
+        Args:
+            p (float): Generalized momentum
```
OleguerCanal: can you state the units (e.g. radiants)

OleguerCanal (changes requested):

environments/environments.py
```diff
@@ -0,0 +1,65 @@
...
+    @abstractmethod
+    def draw(self):
+        """Returns Array of the environment state
+
+        Raises:
+            NotImplementedError: Class instantiation has no implementation
+        """
+        raise NotImplementedError
+
+    def generate_data(self, total_seconds, fps, dt=0.01):
```
OleguerCanal: Add directory argument to store the dataset
e.g: save_dir = None
If save_dir is not none then save the dataset in that directory

environments/environments.py
```diff
@@ -0,0 +1,65 @@
...
+
+class Environment(ABC):
+
+    @abstractmethod
+    def set(self, p, q):
+        """Sets initial conditions for physical system
+
+        Args:
+            p (float): generalized momentum
+            q (float): generalized position
```
fedetask: Could we have `p` and `q` as numpy arrays, so it's more general?

environments/environments.py
```diff
@@ -0,0 +1,65 @@
...
+
+        Raises:
+            NotImplementedError: Class instantiation has no implementation
+        """
+        raise NotImplementedError
```
fedetask, line 26: If I'm not mistaken, a class that extends `ABC` throws an error if it's instantiated without implementation of methods decorated with `@abstractmethod`. Thus, it should not be necessary to raise the `NotImplementedError` in all these methods:

```suggestion
        """
        pass
```

fedetask: Moreover, if someone tries to call `super().set()` (which is something kinda common) in the implementation of the derived class, then an error is raised

environments/pendulum.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    def __init__(self, mass, length, p=None, q=None):
+        """Contructor for pendulum system
+
+        Args:
+            mass (float): Pendulum mass
+            length (float): Pendulum length
+            p (float, optional): Generalized momentum. Defaults to None.
+            q ([type], optional): Generalized position. Defaults to None.
```
fedetask: I am not sure I understand the description here: in what sense `p` and `q` are generalized?

environments/environments.py
```diff
@@ -0,0 +1,65 @@
...
+        phase_state_list = [np.array([self.q, self.p])]
+
+        for _ in range(total_images):
+            current_time = 0
+            while(current_time < time_evol):
+                self.step(dt)
+                current_time += dt
+            image_list.append(self.draw())
+            phase_state_list.append(np.array([self.q, self.p]))
+        return {'input': np.array(image_list), 'groundtruth': np.array(phase_state_list)}
```
fedetask: If we think of the environment as an isolated thing, I would just call the outputs something like `frames` and `states`, since `inputs` and `groundtruth` are referred to the Hamiltonian network and don't have a clear meaning outside that context.

fedetask (commented):
Very good job! I like the structure of the classes and the code. I made some comments on things that I didn't fully understand. Can you check them and tell me what you think?

environments/pendulum.py
fedetask: Ok I understood what they mean here: [Generalized Coordinates](https://en.wikipedia.org/wiki/Generalized_coordinates). I didn't know about this. Should we call them generalized momentum and position only in the abstract `Env` class? Here they represent the cartesian position and the momentum as `m * v`, right?

environments/environments.py
charlio23: Yes Indeed!

environments/pendulum.py
charlio23: In the case of the pendulum, the generalized coordinates correspond to the phase and angular momentum of the ball

environments/environments.py
charlio23: So then, I remove all the raise NotImplementedError in the abstract class?

environments/environments.py
charlio23: Agreed

environments/environments.py
fedetask: Yes, you can substitute them with just `pass`.

environments/spring.py
```diff
@@ -19,20 +19,25 @@ def __init__(self, mass, elastic_cst, p=None, q=None):
...
-            q (float): Generalized position
+            p ([float]): Generalized momentum in 1-D space: Linear momentum
+            q ([float]): Generalized position in 1-D space: Position
+        
+        Raises:
+            ValueError: If p and q are not in 1-D space
         """
+        if len(p) != 1 or len(q) != 1:
+            raise ValueError("p and q must be in 1-D space: Linear momentum and Position.")
         self.p = p
```
OleguerCanal: Maybe it would have been easier to save self.p = p[0] instead of using self.p[0] everywhere else

environments/environments.py
charlio23: depending on the system this does not apply, for a pendulum its the length, for a spring its the elastic constant, for gravity it is the G constant, etc

environments/environments.py
```diff
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 import numpy as np
-
+import os
```
OleguerCanal, line 2: Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.

And  within this groups in alphabetical order

environments/pendulum.py
fedetask: Yes, my bad

environments/environments.py
charlio23: I need more work on this since we will have a N-D system with M objects

environments/environments.py
```diff
@@ -38,28 +38,34 @@ def draw(self):
...
 
         Args:
             total_seconds (int): Total duration of video (in seconds)
             fps (int): Frame rate of generated data (frames per second)
+            save_dir (string, optional): If not None then save the dataset in directory. Defaults to None.
             dt (float, optional): Time step run for the integration. Defaults to 0.01.
 
         Returns:
-            (dict): Contains input data and corresponding ground truth phase states
+            (dict): Contains frames and corresponding phase states
```
OleguerCanal: What are the dict keys?

environments/pendulum.py
```diff
@@ -19,8 +19,8 @@ def __init__(self, mass, length, p=None, q=None):
         Args:
             mass (float): Pendulum mass
             length (float): Pendulum length
-            p (float, optional): Generalized momentum. Defaults to None.
-            q ([type], optional): Generalized position. Defaults to None.
+            p ([float], optional): Generalized momentum in 1-D space: Angular momentum. Defaults to None
```
OleguerCanal: Units?

environments/pendulum.py
```diff
@@ -30,9 +30,14 @@ def set(self, p, q):
         """Sets initial conditions for pendulum
 
         Args:
-            p (float): Generalized momentum
-            q (float): Generalized position
+            p ([float]): Generalized momentum in 1-D space: Angular momentum
```
OleguerCanal: Missing units

environments/spring.py
```diff
@@ -19,20 +19,25 @@ def __init__(self, mass, elastic_cst, p=None, q=None):
...
+            p ([float], optional): Generalized momentum in 1-D space: Linear momentum. Defaults to None
+            q ([float], optional): Generalized position in 1-D space: Position. Defaults to None
         """
         self.mass = mass
         self.elastic_cst = elastic_cst
         self.set(p, q)
 
     def set(self, p, q):
-        """Sets initial conditions for spring
+        """Sets initial conditions for pendulum
```
OleguerCanal: Shouldnt this be spring?

environments/pendulum.py
```diff
@@ -62,8 +67,8 @@ def draw(self):
         draw = ImageDraw.Draw(img)
```
OleguerCanal: Should the image shape be a parameter?

environments/pendulum.py
```diff
@@ -30,9 +30,14 @@ def set(self, p, q):
...
         Args:
-            p (float): Generalized momentum
-            q (float): Generalized position
+            p ([float]): Generalized momentum in 1-D space: Angular momentum
+            q ([float]): Generalized position in 1-D space: Phase
+        
+        Raises:
+            ValueError: If p and q are not in 1-D space
         """
+        if len(p) != 1 or len(q) != 1:
```
OleguerCanal, line 42: Didnt try it but doesnt this crash if you instantiates the pendulum class without specifying p,q? Set is called with None,None

environments/spring.py
```diff
@@ -19,20 +19,25 @@ def __init__(self, mass, elastic_cst, p=None, q=None):
...
             mass (float): Spring mass
             elastic_cst (float): Spring elastic constant
-            p (float, optional): Generalized momentum. Defaults to None.
-            q ([type], optional): Generalized position. Defaults to None.
+            p ([float], optional): Generalized momentum in 1-D space: Linear momentum. Defaults to None
+            q ([float], optional): Generalized position in 1-D space: Position. Defaults to None
         """
         self.mass = mass
         self.elastic_cst = elastic_cst
         self.set(p, q)
```
OleguerCanal: Same as before, I think this will crash if p and q are not passed

OleguerCanal (changes requested):

environments/environments.py
OleguerCanal: I meant the set(p, q) line

environments/spring.py
charlio23: self.q and self.p must be vectors because the evolution is implemented in the abstract class and they are treated in an arbitrary space

environments/environments.py
charlio23: I changed the order now

environments/spring.py
charlio23: Ignore spring implementation in this PR

environments/pendulum.py
charlio23: Changed image visualization method

charlio23:
Merge this PR as soon as you can!

environments/environments.py
OleguerCanal: os is from standard library. Order should be:
```python
from abc import ABC, abstractmethod
import os

import numpy as np
from scipy.integrate import solve_ivp
```

environments/pendulum.py
```diff
@@ -1,5 +1,5 @@
 from .environments import Environment
```
OleguerCanal: This is a local application/library specific imports.
Should be at the end

environments/pendulum.py
```diff
@@ -18,30 +18,31 @@ def __init__(self, mass, length, g, p=None, q=None):
...
-            q ([float]): Generalized position in 1-D space: Phase
-        
+            p ([float]): Generalized momentum in 1-D space: Angular momentum (kg*m^2/s)
+            q ([float]): Generalized position in 1-D space: Phase (rad)
+
         Raises:
             ValueError: If p and q are not in 1-D space
         """
+        
         if len(p) != 1 or len(q) != 1:
```
OleguerCanal, line 42: Maybe also check if they are NOne

environments/pendulum.py
```diff
@@ -79,21 +79,24 @@ def draw(self, res=32, color=True):
         Returns:
```
OleguerCanal, line 64: You need to say evolution() need to be called first so that self.rollout is populated. Maybe this should be a protected function?

environments/pendulum.py
```diff
@@ -1,5 +1,5 @@
 from .environments import Environment
-from PIL import Image, ImageDraw
+
 import numpy as np
 from skimage.draw import circle
```
OleguerCanal: Add scikit-image to requirements.txt

environments/pendulum.py
```diff
@@ -79,21 +79,24 @@ def draw(self, res=32, color=True):
...
-        q = self.rollout[0,:]
+        q = self.rollout[0, :]
         length = len(q)
         if color:
             vid = np.zeros((length, res, res, 3), dtype='float')
         else:
             vid = np.zeros((length, res, res, 1), dtype='float')
         SIZE = 1.5
-        grid = np.arange(0, 1, 1. / res) *2*SIZE  - SIZE
+        grid = np.arange(0, 1, 1. / res) * 2*SIZE - SIZE
```
OleguerCanal: Use autopep

OleguerCanal (changes requested):
Very good job!

General comments:
1. Think about what methods should be protected/private: I think there is some functionality overlapping and it is a bit confusing because you need to rely on some methods side-effects (e.g self.rollout in evolution) in so that others work (draw). I think `evolution` and `dynamics` should be protected. And `generate_data` and `sample_random_rollouts` should be merged into one function (they do almost the same).

2. Would be nice to have some example code which sets up the pendulum environment and shows a visualization to check that everything works and see the work-flow.

Maybe add in the pendulum script something like:
```python
if __name__ == "__main__":
    env = Pendulum(mass=1, length=1, g=10)
    env.sample_init_conditions(radius=5)
    rollout = env.generate_data(total_time=5, delta_time=1)
    visualize(rollout)
```

3. `sample_init_conditions` and `dynamics` in the abstract class should be marked as abstract methods

4. is the `step` function still being used? If not delete it.

5. Delete Spring code from this PR if you dont want it to be merged

environments/pendulum.py
charlio23: It is done in the constructor. I will move it here.

environments/environments.py
charlio23: :+1:

charlio23:
PR ready

Issue #9, opened 2020-09-29
Doubts


OleguerCanal:
I don't fully understand the following things from the paper:
**Implementation**
- [ ] In the re-parametrization trick part, do we estimate the **means** and **variances** of z distribution or just **means** and assume 1 std dev? If estimating both, do they have a 2-headed encoder, one head for means and one for variances?

**Method**
- [ ] What if the total system energy is not constant? (e.g. friction)

fedetask:
Other things that are unclear:
**Implementation**
- [ ] How is the result of Hamiltonian network convolutional layers turned into a scalar? Should we use a linear layer at the end?
- [ ] What is the shape of the latent encoding z? The paper says it has 48 channels but doesn't say anything about the H and W dimensions. Should we assume it is 32x32?

fedetask:
- [ ] Does the decoder need a final sigmoid activation? I would say so since we want values to always lie in [0, 1]. But maybe this goes against the use of L2 loss.

PR #11, opened 2020-10-13
Networks

### LINKED ISSUE #2: Architectures
I was thinking it would be good to have the three different architectures the model combines in different scripts so things are a bit more organized.

Maybe it would be good to have an "architectures" folder and inside there a script for each module. Then a script which combines all of them as a single model.

What do you think @fedetask ?

fedetask:
Adding network models

- Inference networks (encoder and decoder)
- Hamiltonian network
- Decoder network (with residual blocks)

Please, if you can give particular attention to these:

- Activations in encoder, transformer convolutional layers. The paper doesn't say anything about that.
- I haven't checked yet if `concat_rgb()` in `networks/inference_net.py` works as intended. Tensor shape is correct though.
- Should we use batch normalization in residual blocks? The paper doesn't say anything about it but I read that in ResNet it is really helpful.


networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        super().__init__()
+        padding = int(kernel_size / 2)
+        self.input_conv = nn.Conv2d(
+            in_channels=seq_len * img_channels,
+            out_channels=32,
+            kernel_size=3,
+            padding=1
+        )
+        self.conv1 = nn.Conv2d(in_channels=32, out_channels=64, kernel_size=3, padding=padding)
```
OleguerCanal: Maybe if they don't specify kernel sizes, make them a configurable parameter so we can test different values more easily.

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+
+        Args:
+            seq_len (int): Number of frames that compose a sequence.
+            img_channels (int): Number of channels of images in the sequence.
+            out_channels (int): Number of in_channels of the latent encoding.
+            kernel_size (int): Size of kernel to use in the convolutional layers.
+            dtype (torch.dtype): Type of the weights.
+        """
+        super().__init__()
+        padding = int(kernel_size / 2)
```
OleguerCanal: Make it a parameter

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+    The network expects the RGB values to be concatenated. This means that if image shape is
+    (3, 32, 32) and we have 10 images in a sequence, the network will accept an input of shape
+    (N, 3 * 10, 32, 32), where N is the batch size.
+    """
+
+    def __init__(self, seq_len, img_channels, out_channels, kernel_size=3, dtype=torch.float):
+        """Instantiate the 8 convolutional layers that compose the input network with the
+        appropriate shapes.
+
+        TODO: Should we use activations such as ReLU or LeakyReLU?
```
OleguerCanal: I think so, we can use relu for now

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+            padding=padding
+        )
+        self.type(dtype)
+
+    def forward(self, x):
+        """Compute the encoding of the given sequence of images.
+
+        Args:
+            x (torch.Tensor): A N x S x H x W tensor containing the sequence of frames. N is the
+                batch size, S the number of frames in a sequence, H and W the height and width.
```
OleguerCanal: S the number of frames times the number of channels

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+        super().__init__()
+        padding = int(kernel_size / 2)
+        self.input_conv = nn.Conv2d(
+            in_channels=seq_len * img_channels,
+            out_channels=32,
+            kernel_size=3,
+            padding=1
+        )
+        self.conv1 = nn.Conv2d(in_channels=32, out_channels=64, kernel_size=3, padding=padding)
+        self.conv2 = nn.Conv2d(in_channels=64, out_channels=64, kernel_size=3, padding=padding)
```
OleguerCanal: Maybe for testing different architectures it would have been easier to take a list of layer properties:
` [(channels_layer1, kernel_size_layer_1), ..., (channels_layer_n, kernel_size_layer_n)`

And use it to build the architecture, not sure if we need it for now though

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+        z = mean + stddev * epsilon
+        return z, mean, stddev
+
+
+class TransformerNet(nn.Module):
+    """Implementation of the encoder-transformer network, that maps the latent space into the
+    phase space.
+    """
+
+    def __init__(self, in_channels, out_channels, dtype=torch.float):
```
OleguerCanal: Same comments as in the previous class, I think its better if we take more things as parameters to make the tests easier.

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+        self.conv1 = nn.Conv2d(
+            in_channels=in_channels,
+            out_channels=64,
+            kernel_size=3,
+            padding=1,
+            stride=2
+        )
+        self.conv2 = nn.Conv2d(in_channels=64, out_channels=64, kernel_size=3, padding=1, stride=2)
+        self.conv3 = nn.Conv2d(in_channels=64, out_channels=64, kernel_size=3, padding=1, stride=2)
+        self.out_conv = nn.Conv2d(
```
OleguerCanal, line 200: The paper says they use 3 Conv Layers, not 4.

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+            in_channels=64,
+            out_channels=out_channels,
+            kernel_size=3,
+            padding=1
+        )
+        self.type(dtype)
+
+    def forward(self, x):
```
OleguerCanal, line 210: No docstring

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+        Args:
+            x (torch.Tensor): A N x S x H x W tensor containing the sequence of frames. N is the
+                batch size, S the number of frames in a sequence, H and W the height and width.
+
+        Returns:
+            A tuple z, mu, stddev, that are all N x 48 x H x W tensors. z is the latent encoding
+            for the given input sequence, while mu and stddev are the parameters of the
+            variational distribution.
+        """
+        x = self.input_conv(x)
```
OleguerCanal: Again, we should probably use activations

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+    half_len = int(encoding.shape[1] / 2)
+    q = encoding[:, :half_len]
+    p = encoding[:, half_len:]
+    return q, p
+
+
+def concat_rgb(sequence):
+    """Concatenate the images along channel dimension.
+
+    TODO: Is this correct? Need to check!
```
OleguerCanal: Can you test it with simple examples before merging to master?

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+        q and p, that are (N, C/2, ...) tensors.
+    """
+    assert encoding.shape[1] % 2 == 0, 'The number of in_channels is odd. Cannot split into q and p.'
+    half_len = int(encoding.shape[1] / 2)
+    q = encoding[:, :half_len]
+    p = encoding[:, half_len:]
+    return q, p
+
+
+def concat_rgb(sequence):
```
OleguerCanal: Where is this function called? Does it make sense that it is in this script?

networks/inference_net.py
```diff
@@ -0,0 +1,155 @@
...
+    def forward(self, x):
+        x = self.conv1(x)
+        x = self.conv2(x)
+        x = self.conv3(x)
+        x = self.out_conv(x)
+        q, p = to_phase_space(x)
+        return q, p
+
+
+def to_phase_space(encoding):
```
OleguerCanal: Should this be a private method of the transformer class? Or is it called other places?

networks/hamiltonian_net.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, in_channels, dtype=torch.float):
+        """Create the layers of the Hamiltonian network.
+
+        Args:
+            in_channels (int): Number of input in_channels.
+            dtype (torch.dtype): Type of the weights.
+        """
+        super().__init__()
+        self.num_flat_features = 64 * 4 * 4
+        self.in_conv = nn.Conv2d(in_channels=in_channels, out_channels=64, kernel_size=3, padding=1)
```
OleguerCanal: Again, I would like the hardcoded parameters and number of layers to be arguments.

I would like the architecture to be easy to define from a yaml file so that testing different architectures becomes very easy. Maybe we dont need it for now but I think it would be very good to have for when we need to experiment.

networks/hamiltonian_net.py
```diff
@@ -0,0 +1,56 @@
...
+        self.conv1 = nn.Conv2d(in_channels=64, out_channels=64, kernel_size=3, padding=1)
+        self.conv2 = nn.Conv2d(in_channels=64, out_channels=64, kernel_size=3, padding=1)
+        self.conv3 = nn.Conv2d(in_channels=64, out_channels=64, kernel_size=3, padding=1)
+        self.conv4 = nn.Conv2d(in_channels=64, out_channels=64, kernel_size=3, padding=1)
+        self.out_conv = nn.Conv2d(in_channels=64, out_channels=64, kernel_size=3, padding=1)
+        self.linear = nn.Linear(in_features=self.num_flat_features, out_features=1)
+        self.type(dtype)
```
OleguerCanal, line 89: Is this enough to make all the parameters be of dtype?

networks/hamiltonian_net.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def forward(self, q, p):
+        """Forward pass that returns the Hamiltonian for the given q and p inputs.
+
+        q and p must be two N x C x H x W tensors, where N is the batch size, C the number of
+        in_channels, H and W the height and width.
+
+        TODO: Should we use activations such as ReLU or LeakyReLU?
```
OleguerCanal: As we discussed earlier, lets use relu for now after each COnv layer

networks/hamiltonian_net.py
```diff
@@ -0,0 +1,56 @@
...
+
+        Args:
+            q (torch.Tensor): The tensor corresponding to the position in abstract space.
+            p (torch.Tensor): The tensor corresponding to the momentum in abstract space.
+
+        Returns:
+            A N x 1 tensor with the Hamiltonian for each input in the batch.
```
OleguerCanal: The energy

networks/decoder_net.py
```diff
@@ -0,0 +1,130 @@
...
+        Args:
+            x (torch.Tensor): Tensor of shape (N, in_channels, H, W) where N is the batch size.
+
+        Returns:
+            Tensor of shape (out_channels, H * 2^3, W * 2^3) with the reconstructed image.
+        """
+        x = self.res1(x)
+        x = self.res2(x)
+        x = self.res3(x)
+        x = self.sigmoid(self.out(x))
```
OleguerCanal: They dont say anything about sigmoid activation in the lasc Conv layer... Did you get it from karras et al 2018?

networks/decoder_net.py
```diff
@@ -0,0 +1,130 @@
...
+        elif padding == 'same':
+            use_padding = int(kernel_size / 2)
+        else:
+            raise ValueError('Padding argument not understood. Must be integer or "same".')
+        self.conv1 = nn.Conv2d(
+            in_channels=in_channels,
+            out_channels=n_filters,
+            kernel_size=kernel_size,
+            padding=use_padding,
+            bias=False
```
OleguerCanal: No bias?

networks/decoder_net.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Args:
+            x (torch.Tensor): Input image of shape (N, C, H, W) where N is the batch size and C
+                is the number of in_channels.
+
+        Returns:
+            A torch.Tensor with the upsampled images, of shape (N, n_filters, H, W).
+        """
+        x = self.upsample(x)
+        residual = self.dim_match_conv(x) if self.channels != self.n_filters else x
```
OleguerCanal, line 76: Is this a deep copy or a pointer to x?

OleguerCanal (changes requested):
Very nice job! I added some comments of things to look at.

networks/inference_net.py
fedetask: We decided to keep it to automatically adapt in order to maintain the layer input-output size.

networks/decoder_net.py
fedetask: I had copied the code and I forgot to change it. I think we should have biases

networks/decoder_net.py
fedetask: What do you mean? Here I reassign the variable `x` to the upsampled value, and if the number of channels does not match, I store in residual the result of a 1x1 convolution that turns x into a tensor with the correct number of channels.

networks/decoder_net.py
fedetask: No I didn't. You're right, maybe this is wrong. It feels weird that an NN that outputs an image doesn't have a sigmoid activation at the end since we want all outputs to be in [0, 1].

networks/inference_net.py
fedetask: Fixed in commit 7143042. Give a look when you can

networks/inference_net.py
fedetask: I set ReLU as default in commit 7143042

networks/inference_net.py
fedetask: Fixed it in daf5498.

networks/inference_net.py
fedetask: Fixed it in 7143042

networks/inference_net.py
fedetask: Fixed in 7143042

networks/inference_net.py
fedetask: Now the number of hidden layers is parametrized 7143042

networks/inference_net.py
fedetask: Fixed

networks/inference_net.py
fedetask: Done

networks/inference_net.py
fedetask: I think it should be moved into the utilities package, as it will be used to convert the data before passing it to the encoder network

networks/inference_net.py
fedetask: Tested, it is correct. I removed the TODO. Also, I added some tests for the networks

networks/inference_net.py
fedetask: Fixed in 4b0edd6

networks/hamiltonian_net.py
fedetask: Done as we discussed

networks/hamiltonian_net.py
fedetask: Yes it is, I checked in the [docs](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module)

networks/hamiltonian_net.py
fedetask: Done

networks/hamiltonian_net.py
fedetask: Fixed

networks/decoder_net.py
fedetask: I removed it for now. We will check this later. I'm adding it to #9

networks/debug_networks.py
```diff
@@ -0,0 +1,120 @@
+"""This module contains simple, easy to debug networks that have the same interface as the other
+networks in the package.
+"""
+
+import torch
+from torch import nn
+
+from .inference_net import to_phase_space
```
OleguerCanal: Probably this doesnt work any more

networks/inference_net.py
```diff
@@ -0,0 +1,276 @@
...
+    """Implementation of the encoder network, that encodes the input frames sequence into a
+    distribution over the latent space and samples with the common reparametrization trick.
+
+    The network expects the images to be concatenated along channel dimension. This means that if
+    a batch of sequences has shape (batch_size, seq_len, channels, height, width) the network
+    will accept an input of shape (batch_size, seq_len * channels, height, width).
+    """
+
+    DEFAULT_PARAMS = {
+        'hidden_conv_layers': 6,
```
OleguerCanal, line 22: Isn't that number implicit in the len of the lists?

networks/inference_net.py
```diff
@@ -0,0 +1,276 @@
...
+        self.hidden_layers = nn.ModuleList(
+            modules=[
+                nn.Conv2d(
+                    in_channels=n_filters[i],
+                    out_channels=n_filters[i + 1],
+                    kernel_size=kernel_sizes[i + 1],
+                    padding=paddings[i + 1],
+                    stride=strides[i + 1]
+                )
+                for i in range(hidden_conv_layers)
```
OleguerCanal, line 86: nice

networks/hamiltonian_net.py
```diff
@@ -0,0 +1,116 @@
...
+        for i in range(hidden_conv_layers):
+            out_size = int((out_size - kernel_sizes[i] + 2 * paddings[i]) / strides[i]) + 1
+        self.out_conv = nn.Conv2d(
+            in_channels=n_filters[-1],
+            out_channels=n_filters[-1],
+            kernel_size=kernel_sizes[-1],
+            padding=paddings[-1]
+        )
+        out_size = int((out_size - kernel_sizes[-1] + 2 * paddings[-1]) / strides[-1]) + 1
+        self.n_flat = out_size ** 2 * n_filters[-1]
```
OleguerCanal: maybe put parenthesis just in case

networks/hamiltonian_net.py
```diff
@@ -0,0 +1,116 @@
...
+            x = self.activation(layer(x))
+        x = self.activation(self.out_conv(x))
+        x = x.view(-1, self.n_flat)
+        x = self.linear(x)
+        return x
+
+
+if __name__ == '__main__':
+    hamiltonian_net = HamiltonianNet(in_shape=(16, 4, 4))
+    q, p = torch.randn((2, 128, 16, 4, 4))
```
OleguerCanal, line 115: probably im confused, but isn't there an extra dimension here?

networks/decoder_net.py
OleguerCanal: Just wanted to make sure that residual and x where in different things, i guess thats the case sorry

tests/test_inference_net.py
```diff
@@ -0,0 +1,278 @@
+"""This module provides tests for the network architectures. Requires pytest (pip install pytest).
```
OleguerCanal, line 1: Add it to requirements.txt

OleguerCanal (commented):
Very nice updates!!

I added minor comments

networks/debug_networks.py
fedetask: How should we do this?

networks/inference_net.py
fedetask: It is not really intuitive, as the `kernel_sizes` and `strides` lists have len equal to the total number of layers (`hidden_conv_layers + 2`),  while `n_filters` must have len equal to the total number of layers - 1, which is `hidden_conv_layers + 1`.

This is because `strides` and `kernel_sizes` specify the value for each layer of the network. `n_filters` instead doesn't specify the number of filters of the last layer, as it is given by `out_channels`

networks/hamiltonian_net.py
fedetask: You mean `self.n_flat = (out_size ** 2) * n_filters` ?

networks/hamiltonian_net.py
fedetask: It is a dirty trick to avoid writing two lines. The first dimension = 2  is because `q` and `p` are two tensors, so the `(2, 128, 16, 4, 4)` tensor will be split in two and assigned to `q` and `p`. It is equivalent to writing:
```python
q = torch.randn((128, 16, 4, 4))
p = torch.randn((128, 16, 4, 4))
```

networks/decoder_net.py
fedetask: If the number of channels of `x` and the out convolutional layer are the same, then `residual = x`, else `residual` is reshaped to the correct number of channels by a 1x1 convolution.

tests/test_inference_net.py
fedetask: Done in 1bd8362

fedetask (commented):
I answered your comments

networks/hamiltonian_net.py
OleguerCanal: oh i see

networks/hamiltonian_net.py
OleguerCanal: yes

networks/inference_net.py
OleguerCanal: oh i see, ok then

networks/debug_networks.py
OleguerCanal: Not sure... Maybe inherit from the transformer so the class also has it?

fedetask: I just realized that these networks are all not working anymore since I refactored the real ones. I'll make new debug nets that extend the real ones, but do not call their `__init__` and `forward()`.

networks/hamiltonian_net.py
fedetask: Done in 91a0f67

PR #13, opened 2020-10-14
Spring env

### LINKED ISSUE #12: Spring environment
Spring environment implementation.

It should consist on the physical system of a spring attached to an object with mass.

charlio23:
Pull request for spring environment!

### DIFF
--- a/environments/environments.py
+++ b/environments/environments.py
@@ -79,7 +79,7 @@ class Environment(ABC):
         y0 = np.array([np.array(self.q), np.array(self.p)]).reshape(-1)
         self._rollout = solve_ivp(self._dynamics, t_span, y0, t_eval=t_eval).y
 
-    def sample_random_rollouts(self, number_of_frames=100, delta_time=0.1, number_of_rollouts=16, img_size=32, color=True, noisy_data=False, noise_std=0.1, base_radius=1.3, seed=None):
+    def sample_random_rollouts(self, number_of_frames=100, delta_time=0.1, number_of_rollouts=16, img_size=32, color=True, noisy_data=False, noise_std=0.1, radius_lb=1.3, radius_ub=2.3, seed=None):
         """Samples random rollouts for a given environment
 
         Args:
@@ -90,20 +90,25 @@ class Environment(ABC):
             color (bool): Whether to have colored or grayscale frames.
             noisy_data (bool): Whether to add noise to data.
             noise_std (float): If noisy_data=true, standard deviation of the gaussian noise source.
-            base_radius (float): Init phase states will be sampled from a circle (q, p) of radius 
+            radius_lb (float): Radius lower bound of the phase state sampling.
+            radius_ub (float): Radius upper bound of the phase state sampling.
+                Init phase states will be sampled from a circle (q, p) of radius 
                 r ~ U(base_radius, base_radius + 1) https://arxiv.org/pdf/1909.13789.pdf (Sec. 4)
-            seed (int): Seed for reproducibility
-
+            seed (int): Seed for reproducibility.
+        Raises:
+            AssertError: If radius_lb > radius_ub
         Returns:
             (ndarray): Array of shape (Batch, Nframes, Height, Width, Channels).
                 Contains sampled rollouts
         """
+
+        assert radius_lb <= radius_ub
         if seed is not None:
             np.random.seed(seed)
         total_time = number_of_frames*delta_time
         batch_sample = []
         for i in range(number_of_rollouts):
-            radius = np.random.rand() + base_radius
+            radius = np.random.rand()*(radius_ub - radius_lb) + radius_lb
             self.sample_init_conditions(radius)
             self._evolution(total_time, delta_time)
             if noisy_data:
--- a/environments/environments.py
+++ b/environments/environments.py
@@ -93,7 +93,7 @@ class Environment(ABC):
             radius_lb (float): Radius lower bound of the phase state sampling.
             radius_ub (float): Radius upper bound of the phase state sampling.
                 Init phase states will be sampled from a circle (q, p) of radius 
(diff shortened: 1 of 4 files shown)


environments/environments.py
```diff
@@ -84,15 +84,7 @@ def _evolution(self, total_time=10, delta_time=0.1):
...
-    def sample_random_rollouts(self,
-                               number_of_frames=100,
-                               delta_time=0.1,
-                               number_of_rollouts=16,
-                               img_size=32,
-                               color=True,
-                               noise_std=0.,
-                               base_radius=1.3,
-                               seed=None):
+    def sample_random_rollouts(self, number_of_frames=100, delta_time=0.1, number_of_rollouts=16, img_size=32, color=True, noisy_data=False, noise_std=0.1, radius_lb=1.3, radius_ub=2.3, seed=None):
```
OleguerCanal: I think its better if we use autopep, easier to read all the arguments.

Also I removed the noisy data flag, if you dont want noisy data just set the std dev to 0. The idea was to not have redundant arguments

environments/environments.py
```diff
@@ -102,21 +94,26 @@ def sample_random_rollouts(self,
...
             np.random.seed(seed)
         total_time = number_of_frames * delta_time
         batch_sample = []
         for i in range(number_of_rollouts):
-            radius = np.random.rand() + base_radius
-            self._sample_init_conditions(radius)
+            radius = np.random.rand()*(radius_ub - radius_lb) + radius_lb
+            self.sample_init_conditions(radius)
             self._evolution(total_time, delta_time)
             if noise_std > 0.:
```
OleguerCanal, line 119: Notice i did this

environments/environments.py
```diff
@@ -102,21 +94,26 @@ def sample_random_rollouts(self,
             img_size (int): Size of the frames (in pixels)
             color (bool): Whether to have colored or grayscale frames.
             noise_std (float): If noisy_data=true, standard deviation of the gaussian noise source.
-            base_radius (float): Init phase states will be sampled from a circle (q, p) of radius 
-                r ~ U(base_radius, base_radius + 1) https://arxiv.org/pdf/1909.13789.pdf (Sec. 4)
-            seed (int): Seed for reproducibility
-
+            radius_lb (float): Radius lower bound of the phase state sampling.
```
OleguerCanal: maybe a tuple is better rather than 2 parameters?

environments/pendulum.py
```diff
@@ -99,21 +99,19 @@ def _sample_init_conditions(self, radius):
         Args:
             radius (float): Radius of the sampling process
         """
-        states = np.random.rand(2) * 2. - 1
-        states /= np.sqrt((states**2).sum()) * radius
+        states = np.random.rand(2)*2.-1
```
OleguerCanal: autopep

environments/pendulum.py
```diff
@@ -99,21 +99,19 @@ def _sample_init_conditions(self, radius):
...
 if __name__ == "__main__":
 
     pd = Pendulum(mass=.5, length=1, g=3)
-    rolls = pd.sample_random_rollouts(number_of_frames=100,
-                                      delta_time=0.1,
-                                      number_of_rollouts=16,
-                                      img_size=32,
-                                      noise_std=0.,
-                                      seed=23)
+    rolls = pd.sample_random_rollouts(number_of_frames=100, delta_time=0.1,
```
OleguerCanal: autopep

environments/spring.py
```diff
@@ -0,0 +1,114 @@
...
+                                      noisy_data=False, noise_std=0.1,
+                                      radius_lb=0.1, radius_ub=1.0, seed=23)
+    fig = plt.figure()
+    img = []
+    idx = np.random.randint(rolls.shape[0])
+    for im in rolls[idx]:
+        img.append([plt.imshow(im, animated=True)])
+    ani = animation.ArtistAnimation(fig, img, interval=50, blit=True,
+                                    repeat_delay=1000)
+    plt.show()
```
OleguerCanal: Could you move these into a function? So we do not need to write it each time we want to debug something. It will also be useful to check the HGN outputs

environments/spring.py
```diff
@@ -0,0 +1,114 @@
...
+        Returns:
+            vid (np.ndarray): Rendered rollout as a sequence of images
+        """
+        q = self._rollout[0, :]
+        length = len(q)
+        if color:
+            vid = np.zeros((length, res, res, 3), dtype='float')
+        else:
+            vid = np.zeros((length, res, res, 1), dtype='float')
+        SIZE = 1.5
```
OleguerCanal: this should be a parameter

environments/spring.py
```diff
@@ -0,0 +1,114 @@
...
+
+    """Spring System
+
+    Equations of movement are:
+
+        x'' = -(k/m)*x
+
+    """
+
+    def __init__(self, mass, elastic_cst, p=None, q=None):
```
OleguerCanal: I'd like that the order is always q, p. To try avoid integration problems with the rest of the system

environments/spring.py
```diff
@@ -0,0 +1,114 @@
...
+
+        Args:
+            mass (float): Spring mass
+            elastic_cst (float): Spring elastic constant
+            p ([float], optional): Generalized momentum in 1-D space: Linear momentum (kg*m/s). Defaults to None
+            q ([float], optional): Generalized position in 1-D space: Position (m). Defaults to None
+        """
+        self.mass = mass
+        self.elastic_cst = elastic_cst
+        super().__init__(p, q)
```
OleguerCanal: Its better to do `(q=q, p=p)` to make sure we don't do mistakes with the position and momentum

OleguerCanal (changes requested):
Very nice!

In a general note: try to run autopep before pushing things so that the formatting is good.

environments/environments.py
charlio23: tuple better!

PR #15, opened 2020-10-14
Gravity env: N bodies

### LINKED ISSUE #14: N-body environment
N-body gravitational environment implementation.

It should consist on the physical system of a gravitation system of N bodies. It should work for 2 and 3 bodies for reproducibility purposes.

Future work may include 4 or more bodies.

charlio23:
Pull request that includes gravity interaction with 2 bodies.

The implementation requires changes in the abstract class, therefore the rest of the environments are refactored to adapt to these changes.

### DIFF
--- a/environments/environments.py
+++ b/environments/environments.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 import os
 
+from matplotlib import pyplot as plt, animation
 import numpy as np
 from scipy.integrate import solve_ivp
 
@@ -53,11 +54,11 @@ class Environment(ABC):
         """
         raise NotImplementedError
 
-    def _sample_init_conditions(self, radius):
+    def _sample_init_conditions(self, radius_bound):
         """Samples random initial conditions for the environment
 
         Args:
-            radius (float): Radius of the sampling process
+            radius_bound (float, float): Radius lower and upper bound of the phase state sampling.
 
         Raises:
             NotImplementedError: Class instantiation has no implementation
@@ -75,8 +76,12 @@ class Environment(ABC):
         Raises:
             AssertError: If p or q are None
         """
-        assert self.q != None
-        assert self.p != None
+        if isinstance(self.q, np.ndarray):
+            assert self.q.all() != None
+            assert self.p.all() != None
+        else:
+            assert self.q != None
+            assert self.p != None
 
         t_eval = np.linspace(0, total_time,
                              round(total_time / delta_time) + 1)[:-1]
@@ -84,40 +89,55 @@ class Environment(ABC):
         y0 = np.array([np.array(self.q), np.array(self.p)]).reshape(-1)
         self._rollout = solve_ivp(self._dynamics, t_span, y0, t_eval=t_eval).y
 
-    def sample_random_rollouts(self, number_of_frames=100, delta_time=0.1, number_of_rollouts=16, img_size=32, color=True, noisy_data=False, noise_std=0.1, radius_lb=1.3, radius_ub=2.3, seed=None):
+    def sample_random_rollouts(self, number_of_frames=100, delta_time=0.1, number_of_rollouts=16, img_size=32, color=True, noise_std=0.1, radius_bound=(1.3, 2.3), world_size=1.5, seed=None):
         """Samples random rollouts for a given environment
 
         Args:
-            number_of_frames (int): Total duration of video (in frames)
-            delta_time (float): Frame interval of generated data (in seconds)
+            number_of_frames (int): Total duration of video (in frames).
+            delta_time (float): Frame interval of generated data (in seconds).
             number_of_rollouts (int): Number of rollouts to generate.
-            img_size (int): Size of the frames (in pixels)
+            img_size (int): Size of the frames (in pixels).
             color (bool): Whether to have colored or grayscale frames.
-            noise_std (float): If noisy_data=true, standard deviation of the gaussian noise source.
-            radius_lb (float): Radius lower bound of the phase state sampling.
-            radius_ub (float): Radius upper bound of the phase state sampling.
-                Init phase states will be sampled from a circle (q, p) of radius 
-                r ~ U(radius_lb, radius_ub) https://arxiv.org/pdf/1909.13789.pdf (Sec. 4)
+            noise_std (float): Standard deviation of the gaussian noise source, no noise for noise_std=0.
+            radius_bound (float, float): Radius lower and upper bound of the phase state sampling.
+                Init phase states will be sampled from a circle (q, p) of radius
+                r ~ U(radius_bound[0], radius_bound[1]) https://arxiv.org/pdf/1909.13789.pdf (Sec. 4)
+            world_size (float) Spatial extent of the window where the rendering is taking place (in meters).
             seed (int): Seed for reproducibility.
         Raises:
-            AssertError: If radius_lb > radius_ub
+            AssertError: If radius_bound[0] > radius_bound[1]
         Returns:
             (ndarray): Array of shape (Batch, Nframes, Height, Width, Channels).
                 Contains sampled rollouts
         """
-
+        radius_lb, radius_ub = radius_bound
         assert radius_lb <= radius_ub
         if seed is not None:
             np.random.seed(seed)
         total_time = number_of_frames * delta_time
         batch_sample = []
         for i in range(number_of_rollouts):
-            radius = np.random.rand()*(radius_ub - radius_lb) + radius_lb
-            self.sample_init_conditions(radius)
+            self._sample_init_conditions(radius_bound)
             self._evolution(total_time, delta_time)
             if noise_std > 0.:
                 self._rollout += np.random.randn(
                     *self._rollout.shape) * noise_std
-            batch_sample.append(self._draw(img_size, color))
+            batch_sample.append(self._draw(img_size, color, world_size))
 
         return np.array(batch_sample)
+
+
+def visualize_rollout(rollout):
+    """Visualization for a single sample rollout of a physical system.
+
+    Args:
+        rollout (ndarray): Rollout of an environment.
+            Its shape should be (Nframes, Height, Width, Channels)
+    """
+    fig = plt.figure()
+    img = []
+    for im in rollout:
+        img.append([plt.imshow(im, animated=True)])
+        ani = animation.ArtistAnimation(fig, img, interval=50,
+                                        blit=True, repeat_delay=1000)
+    plt.show()
--- a/environments/gravity.py
+++ b/environments/gravity.py
@@ -13,71 +13,38 @@ class NObjectGravity(Environment):
 
     """
 
-    def __init__(self, mass, gravity_cst, p=None, q=None):
+    def __init__(self, mass, gravity_cst, q=None, p=None):
         """Contructor for spring system
 
         Args:
             mass ([float]): Object masses (kg).
             gravity_cst (float): Constant for the intensity of gravitational field (m^3/kg*s^2)
-            p (ndarray, optional): Object generalized momentums in 2-D space : Linear momentums (kg*m/s). Defaults to None
             q (ndarray, optional): Object generalized positions in 2-D space: Positions (m). Defaults to None
+            p (ndarray, optional): Object generalized momentums in 2-D space : Linear momentums (kg*m/s). Defaults to None
         """
         self.mass = mass
         self.n_objects = len(mass)
         self.gravity_cst = gravity_cst
-        super().__init__(p, q)
+        super().__init__(q=q, p=p)
 
-    def set(self, p, q):
+    def set(self, q, p):
         """Sets initial conditions for pendulum
 
         Args:
-            p (ndarray): Object generalized momentums in 2-D space : Linear momentums (kg*m/s)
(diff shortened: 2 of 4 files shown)


charlio23:
Updated title, now it works for up to 7 bodies.

environments/environments.py
```diff
@@ -54,11 +54,11 @@ def _draw(self):
         """
         raise NotImplementedError
 
-    def _sample_init_conditions(self, radius):
+    def _sample_init_conditions(self, radius_bound):
         """Samples random initial conditions for the environment
 
         Args:
-            radius (float): Radius of the sampling process
+            radius_bound (float, float): Radius lower and upper bound of the phase state sampling.
```
fedetask: Since `radius_bound` is a tuple, should the docstring be something like
```suggestion
            radius_bound (tuple): Tuple with radius lower and upper bound of the phase state sampling.
```

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+
+        m_i*q'' = G * sum_{i!=j} ( m_i*m_j*(q_j - q_i)/ abs(q_j  - q_i)^3 )
+
+    """
+
+    def __init__(self, mass, gravity_cst, orbit_noise=.01, q=None, p=None):
+        """Contructor for spring system
+
+        Args:
+            mass ([float]): Object masses (kg).
```
fedetask: As above, should the docstring be
```suggestion
            mass (list): List of floats corresponding to object masses (kg).
```

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+        if q is None or p is None:
+            return
+        if q.shape[0] != self.n_objects or p.shape[0] != self.n_objects:
+            raise ValueError(
+                "q and p do not refer to the same number of objects in the system.")
+        if q.shape[-1] != 2 or p.shape[-1] != 2:
+            raise ValueError(
+                "q and p must be in 2-D space: Position and Linear momentum.")
+        self.q = q.copy()
+        self.p = p.copy()
```
fedetask, line 57: I was wondering: I see that in the environments `self.q` and `self.q` are never actually used for something. What is their purpose? Because my feeling is that they are in the abstract `Environment` class to be then used differently for each environment implementation. But I see that environments take their parameters differently. Maybe we can just remove the whole `q` and `p` thing?

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def _dynamics(self, t, states):
+        """Defines system dynamics
+
+        Args:
+            t (float): Time parameter of the dynamic equations.
+            states ([float]) Phase states at time t
```
fedetask: I'm not sure I understand what `states` is made of. I understand it needs to contain position and momentum information for each body, but in what format?

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+        dyn[0, :, :] = states_p/(np.array(self.mass)[:, np.newaxis])
+
+        # Distance calculation
+        object_distance = np.zeros((self.n_objects, self.n_objects))
+        for i in range(self.n_objects):
+            for j in range(i, self.n_objects):
+                object_distance[i, j] = np.linalg.norm(
+                    states_q[i] - states_q[j])
+                object_distance[j, i] = object_distance[i, j]
+        object_distance = np.power(object_distance, 3)/self.gravity_cst
```
fedetask, line 84: What does `object_distance` become after this line?

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def _dynamics(self, t, states):
+        """Defines system dynamics
+
+        Args:
+            t (float): Time parameter of the dynamic equations.
+            states ([float]) Phase states at time t
+
+        Returns:
+            equations ([float]): Movement equations of the physical system
```
fedetask: Also here, what is actually `equations`? What does each element of this list/array represent?

environments/environments.py
fedetask: Another question: why is the name radius? Can it be something more general?

environments/environments.py
```diff
@@ -54,11 +54,11 @@ def _draw(self):
         """
         raise NotImplementedError
```
fedetask: Not necessary to raise a `NotImplementedError` in `abstractmethod`s. The `ABC` superclass will already raise the correct exception by itself if a class that does not implement the method is instantiated. You can just use pass
```suggestion
        pass
```

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+        self.mass = mass
+        self.colors = ['y', 'r', 'g', 'b', 'c', 'p', 'w']
+        self.n_objects = len(mass)
+        self.gravity_cst = gravity_cst
+        self.orbit_noise = orbit_noise
+        if self.n_objects > 7:
+            raise NotImplementedError(
+                'Gravity interaction for ' + str(self.n_objects) + ' bodies is not implemented.')
+        super().__init__(q=q, p=p)
```
fedetask, line 36: What's the use of `q` and `p` here? Because they're always set to `None` in the `super().__init__(q, p)`

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+        """Contructor for spring system
+
+        Args:
+            mass ([float]): Object masses (kg).
+            gravity_cst (float): Constant for the intensity of gravitational field (m^3/kg*s^2)
+            orbit_noise (float, optional): Noise for object orbits when sampling initial conditions
+            q (ndarray, optional): Object generalized positions in 2-D space: Positions (m). Defaults to None
+            p (ndarray, optional): Object generalized momentums in 2-D space : Linear momentums (kg*m/s). Defaults to None
+        Raises:
+            NotImplementedError: If more than 2 objects are considered
```
fedetask:
```suggestion
            NotImplementedError: If more than 7 objects are considered
```

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+        object_distance = np.power(object_distance, 3)/self.gravity_cst
+
+        for d in range(2):
+            for i in range(self.n_objects):
+                mom_term = 0
+                for j in range(self.n_objects):
+                    if i != j:
+                        mom_term += self.mass[j]*(states_q[j, d] -
+                                                  states_q[i, d])/object_distance[i, j]
+                dyn[1, i, d] += mom_term*self.mass[i]
```
fedetask, line 93: Could you add one comment line that briefly explains what you're doing here?

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+    def _draw(self, res=32, color=True, world_size=1.5):
+        """Returns array of the environment evolution
+
+        Args:
+            res (int): Image resolution (images are square).
+            color (bool): True if RGB, false if grayscale.
+            world_size (float) Spatial extent of the window where the rendering is taking place (in meters).
+
+        Returns:
+            vid (np.ndarray): Rendered rollout as a sequence of images
```
fedetask: Just to make it clear the shape of returned things.
```suggestion
            vid (np.ndarray): Numpy array of shape (seq_len, height, width, channels)
                containing the rendered rollout as a sequence of images.
```

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+        q = self._rollout.reshape(2, self.n_objects, 2, -1)[0, :, :, :]
```
fedetask: I think it works also if you don't specify all the dimensions
```suggestion
        q = self._rollout.reshape(2, self.n_objects, 2, -1)[0]
```

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+        length = q.shape[-1]
+        if color:
+            vid = np.zeros((length, res, res, 3), dtype='float')
+        else:
+            vid = np.zeros((length, res, res, 1), dtype='float')
```
fedetask, line 113: There is an ongoing discussion about the shape of this in Issue #26, I'll add a comment there

environments/gravity.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # velocity that yields a circular orbit
+        if self.n_objects == 2:
+            factor = 2
+        else:
+            factor = np.sqrt(np.sin(np.pi/3)/(2*np.cos(np.pi/6)**2))
+        vel = np.flipud(pos)/(factor*radius**1.5)
+        vel[0] *= -1
+        states[0, 0, :] = pos
+        states[1, 0, :] = vel
```
fedetask, line 182: I am a bit worried that this will lead to only very similar trajectories, as they always start in a circle. If the sequence is always taken from this starting state, we will always see only the initial evolution of the system, right? 
Can the initial conditions be less rigid? As long as the bodies don't go outside the boundaries in `seq_len` steps the sequences are valid in my opinion. If they go out, the sequence could be discarded.

fedetask (changes requested):
Good job! My main comments are about
- I am not totally understanding the use of `q` and `p`
- Describe in more detail inputs/outputs of methods in their docstring, particularly the format and shape of arrays.
- Is n-body random initialization diverse enough?

environments/environments.py
charlio23: it is radius because in the paper they call it radius. You can find the information in section 4.

environments/gravity.py
charlio23: The q and p are the initial conditions that are fed in the evolution function used in the abstract class.

So self.p and self.q are just the initial conditions of the physical system.

environments/gravity.py
charlio23: states is a 1-D array that contains the information of the phase state. In a general way, it will be always np.array([q,p]).reshape(-1).

In the N-body environment you can retrieve the phase state array if you reshape it using .reshape(2, self.n_bodies, 2).

The reason why you need the 1-D array is because scipy.solve_ivp requires it.

environments/gravity.py
charlio23: object distance is originally the euclidean distance between two objects |q_j - q_i|. However, the potential energy term of the hamiltonian is g*(q_i - q_j) / |q_i - q_j|^3. Therefore, the object distance has the following form: g / |q_i - q_j|^3.

We can rename it if you wish.

environments/gravity.py
charlio23: `equations ([float]): Movement equations of the physical system`

In other words, the derivative of states wrt time.

environments/environments.py
charlio23: Okay, I just took a placeholder code haha.

environments/gravity.py
charlio23: This is actually a typo in environments.py. I will fix it.

Edit: I checked the code and it is correct. This line describes a use case scenario when the user wants to have some preferred initial conditions for the system when initializing an environment.

environments/gravity.py
charlio23: The problem with the n-body system is that the trajectories are too chaotic for n>2. For the implementation of the initial conditions, I followed the following code
[https://github.com/greydanus/hamiltonian-nn/blob/master/experiment-3body/data.py](url)
I have not checked whether this conditions lead to very constrained trajectories. The idea was to follow the hamiltonian neural network data.

We could try setting less rigid initial conditions but it would be too difficult to find stable trajectories. The previous solution ensures that the trajectory will be stable for 2 and 3 objects. I generalized the approach to make it work for >3 objects, but stability is not guaranteed in these scenarios.

I don't think that discarding sequences is an option if we want to train our models and generate data at the same time.

environments/gravity.py
charlio23: As I commented, I disagree.

The usage of the dataloader is to read the data and prepare it to the networks. The environment is a data generator, not a data pre-processor.

environments/gravity.py
charlio23: I will. These for loops are the computation of the potential energy for each object in each dimension.

environments/environments.py
fedetask: Ok sorry I didn't remember that

environments/gravity.py
fedetask: Sorry I realized only now that the `__init__` of `Environment` is actually setting the variables to none **before** calling `set(q, p)`. Then it makes sense, you're right

environments/gravity.py
fedetask: Ok thanks! can you add that to the docstring? 
```suggestion
            states (numpy.ndarray): 1-D array that contains the information of the phase
                state, in the format of np.array([q,p]).reshape(-1).
```

environments/gravity.py
fedetask: Oh okay. Can it be
```suggestion
            equations (numpy.ndarray): Numpy array with derivatives of q and p w.r.t. time
```
?

environments/gravity.py
fedetask: Ok if they do that in HNN I suppose it is good enough. Let's keep it like this for now

fedetask (changes requested):
Good work! Can you just add comments to those lines we discussed? Then this is ready to be merged.

environments/gravity.py
charlio23: ofc!

charlio23:
ready to merge, need to fix changes before.

PR #17, opened 2020-10-14
Environment Pytorch Dataset

### LINKED ISSUE #16: Pytorch DataLoader class. Data API for feeding the networks
A pytorch dataloader is required to sample data and forward it to the networks.
`class EnvironmentSampler(Dataset):`
The class will have the following structure:

- A constructor: `def __init__(self, environment, length):`
where `environment` is an `Environment` instance to indicate the physical system to draw data from; and `length` indicates the length of the dataset (since we can keep sampling random rollouts we avoid having an infinite size).

- `    def __len__(self):
        return self.length`
- An iterable call `def __getitem__(self, i):`
this will be called each time a batch is required. A rollout is sampled and converted as a pytorch tensor to be forwarded to the network. It should perform standarization of the data (maybe just mean substraction?)

 Sample code will be provided in the same file to show its usage.

charlio23:
EnvironmentSampler Pytorch DataLoader for data sampling.
Samples rollouts from an environment, performs standarization, returns pytorch array in the following format (Batch, Nframes, Channels, Height, Width).

### DIFF
--- a/environments/datasets.py
+++ b/environments/datasets.py
@@ -10,12 +10,12 @@ class EnvironmentSampler(Dataset):
     Given an environment and sampling conditions, the dataset samples rollouts as pytorch tensors.
     """
 
-    def __init__(self, environment, length, number_of_frames, delta_time, number_of_rollouts, img_size, noise_std, radius_bound, world_size, seed, data_mean=.5, data_std=.5):
+    def __init__(self, environment, dataset_len, number_of_frames, delta_time, number_of_rollouts, img_size, noise_std, radius_bound, world_size, seed, data_mean=.5, data_std=.5):
         """Constructor method
 
         Args:
             environment (Environment): Instance belonging to Environment abstract base class.
-            length (int): Length of the dataset.
+            dataset_len (int): Length of the dataset.
             number_of_frames (int): Total duration of video (in frames).
             delta_time (float): Frame interval of generated data (in seconds).
             number_of_rollouts (int): Number of rollouts to generate.
@@ -31,7 +31,7 @@ class EnvironmentSampler(Dataset):
             data_std (float): Data std for standaritzation. Defaults to 0.5.
         """
         self.environment = environment
-        self.length = length
+        self.dataset_len = dataset_len
         self.data_mean = data_mean
         self.data_std = data_std
         self.number_of_frames = number_of_frames
@@ -49,7 +49,7 @@ class EnvironmentSampler(Dataset):
         Returns:
             length (int): Length of the dataset.
         """
-        return self.length
+        return self.dataset_len
 
     def __getitem__(self, i):
         """Iterator for rollout sampling.
@@ -57,6 +57,9 @@ class EnvironmentSampler(Dataset):
 
         Args:
             i (int): Index of the dataset sample (ignored since we sample random data).
+        Returns:
+            (Torch.tensor): Tensor of shape (Batch, Nframes, Channels, Height, Width).
+                Contains standarized sampled rollouts        
         """
         rolls = self.environment.sample_random_rollouts(number_of_frames=self.number_of_frames,
                                                         delta_time=self.delta_time,
--- a/environments/datasets.py
+++ b/environments/datasets.py
@@ -1,0 +1,108 @@
+import torch
+from torch.utils.data import Dataset
+from torchvision import transforms
+
+from pendulum import Pendulum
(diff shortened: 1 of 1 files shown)


environments/datasets.py
```diff
@@ -0,0 +1,108 @@
...
+        return rolls.transpose((0, 1, 4, 2, 3))
+
+
+# Sample code for DataLoader call
+if __name__ == "__main__":
+    import time
+    pd = Pendulum(mass=.5, length=1, g=3)
+    trainDS = EnvironmentSampler(
+        environment=pd,
+        length=100,
```
OleguerCanal, line 83: Dont call it length also, can be confused with the length of the pendulum

environments/datasets.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def __len__(self):
+        """Get dataset length
+
+        Returns:
+            length (int): Length of the dataset.
+        """
+        return self.length
+
+    def __getitem__(self, i):
```
OleguerCanal, line 54: what does it return? Is it a Torch.tensor?

environments/datasets.py
```diff
@@ -0,0 +1,108 @@
...
+            number_of_rollouts (int): Number of rollouts to generate.
+            img_size (int): Size of the frames (in pixels).
+            color (bool): Whether to have colored or grayscale frames.
+            noise_std (float): Standard deviation of the gaussian noise source, no noise for noise_std=0.
+            radius_bound (float, float): Radius lower and upper bound of the phase state sampling.
+                Init phase states will be sampled from a circle (q, p) of radius
+                r ~ U(radius_bound[0], radius_bound[1]) https://arxiv.org/pdf/1909.13789.pdf (Sec. 4)
+            world_size (float) Spatial extent of the window where the rendering is taking place (in meters).
+            seed (int): Seed for reproducibility.
+            data_mean (float): Data mean for standaritzation. Defaults to 0.5.
```
OleguerCanal, line 30: Shouldnt this default to 0 and std dev to 1 to have normalized data?

environments/datasets.py
```diff
@@ -0,0 +1,108 @@
...
+        rolls = self.environment.sample_random_rollouts(number_of_frames=self.number_of_frames,
+                                                        delta_time=self.delta_time,
+                                                        number_of_rollouts=self.number_of_rollouts,
+                                                        img_size=self.img_size,
+                                                        noise_std=self.noise_std,
+                                                        radius_bound=self.radius_bound,
+                                                        world_size=self.world_size,
+                                                        seed=self.seed)
+        # standarization
+        rolls = (rolls - self.data_mean)/self.data_std
```
OleguerCanal, line 73: Arent the rolls from 0-255? I don't understand how you are normalizing here

environments/datasets.py
```diff
@@ -0,0 +1,108 @@
...
+        self.number_of_frames = number_of_frames
+        self.delta_time = delta_time
+        self.number_of_rollouts = number_of_rollouts
+        self.img_size = img_size
+        self.noise_std = noise_std
+        self.radius_bound = radius_bound
+        self.world_size = world_size
+        self.seed = seed
+
+    def __len__(self):
```
OleguerCanal, line 46: is this needed?

OleguerCanal (changes requested):
Nice work! Just minor comments :)

environments/datasets.py
charlio23: If we have an image that is [0...1] to have standarized data [-1 ... 1] we can use 0.5 as mean and std, that is why i took this default value

environments/datasets.py
charlio23: the rolls are [0 ... 1]

environments/datasets.py
charlio23: yes, since EnvironmentSampler class is treated as an iterable in the training loop. I have not tested what happens if this methods is missing but i am sure it won't work without it.

environments/datasets.py
charlio23: yes, i forgot!

charlio23:
I solved the issues. I left some conversations unresolved cuz I am not sure if they answer your doubts.

environments/datasets.py
OleguerCanal: Oh i see, soory. I thought the images were 0-255

PR #18, opened 2020-10-14
Dev oleguer

### LINKED ISSUE #7: Write objective function (loss)

### LINKED ISSUE #10: Integrator
We need to implement the **leapfrog integrator** from the HNN.
We should also implement **euler integration** and **runge-kutta rk4** integrator.

I'm thinking of having an integrator class and pass it to the model. Something like:

```python
integrator = Integrator(method="euler")
q_next, p_next = integrator.step(q_now, p_now, HNN, dt)
```

What do you guys think @charlio23  @fedetask

OleguerCanal:
HGN training pipeline + integrator + loss function

Missing:
- [x] KL divergence in loss
- [x] Saving/Loading of model architecture and params
- [x] Integrating new dataloader (env and models still not connected)


environments/environment_factory.py
```diff
@@ -0,0 +1,59 @@
+"""Environment factory class. Given a valid environment name and its constructor args, returns an instantiation of it
+"""
+from .environments import Environment
+# TODO(Oleguer): Fix this, its not very nice to have to import all classes
+from .pendulum import Pendulum
+# from spring import Spring
```
charlio23: Should we import all the implemented environments here?

environments/pendulum.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, mass, length, g, q=None, p=None):
         self.mass = mass
         self.length = length
         self.g = g
-        super().__init__(q=q, p=p)
+        super(Pendulum, self).__init__(q=q, p=p)
```
charlio23: Why do you call `super `using `Pendulum` and `self`? Does this needs changes in the `Environment` class?

hamiltonian_generative_network.py
```diff
@@ -0,0 +1,122 @@
+import torch
+
+from utilities.integrator import Integrator
+from utilities.hgn_result import HgnResult
+
+
+class HGN():
+    """Hamiltonian Generative Network model.
+    
+    This class models the HGN and allows implements its training and evaluation.
```
charlio23: allows implementing 🍔

hamiltonian_generative_network.py
```diff
@@ -0,0 +1,122 @@
...
+            q, p = self.integrator.step(q=q, p=p, hnn=self.hnn)
+            prediction.append_state(q=q, p=p)
+
+            # Compute state reconstruction
+            x_reconstructed = self.decoder(q=q)
+            prediction.append_reconstruction(x_reconstructed)
+        return prediction
+
+    def fit(self, rollouts):
+        """Perform a training step with the given rollouts batch.
```
charlio23, line 99: at some point in the training step before the forward pass we should have the models uploaded to the GPU as well as the batch that has to be forwarded.

charlio23 (commented):
Nice implementation! Just one important note. In the training script we need to upload the netoworks and the batches to the GPU.

A good practice would be to have a boolean in the yaml file to select if you want to use GPU or CPU computations.

environments/environment_factory.py
OleguerCanal: yes, not ideal

environments/pendulum.py
OleguerCanal: Sorry, I think its not necessary. I'll remove it

hamiltonian_generative_network.py
OleguerCanal: Good point, I'll will ad a check that uses gpu if available

OleguerCanal:
Batch size bigger than one doesnt work (integrator fails), for the rest it seems that the pipeline works without crashing

environments/environment_factory.py
```diff
@@ -0,0 +1,59 @@
+"""Environment factory class. Given a valid environment name and its constructor args, returns an instantiation of it
+"""
+from .environments import Environment
+# TODO(Oleguer): Fix this, its not very nice to have to import all classes
+from .pendulum import Pendulum
+from .spring import Spring
+
```
fedetask: Can we use `sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))`? The relative import is not working for me. 
```suggestion
import sys
import os

sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
from environments import Environment
# TODO(Oleguer): Fix this, its not very nice to have to import all classes
from pendulum import Pendulum
from spring import Spring

```

environments/pendulum.py
```diff
@@ -1,6 +1,6 @@
 import numpy as np
 
-from environments import Environment, visualize_rollout
+from .environments import Environment, visualize_rollout
```
fedetask: Same as above

environments/spring.py
```diff
@@ -1,6 +1,6 @@
 import numpy as np
 
-from environments import Environment, visualize_rollout
+from .environments import Environment, visualize_rollout
```
fedetask: Same as above

hamiltonian_generative_network.py
```diff
@@ -0,0 +1,150 @@
...
+        """Get the prediction of the HGN for a given rollout_batch of n_steps.
+
+        Args:
+            rollout_batch (torch.Tensor(N, C, H, W)): Image sequence of the system evolution concatenated along the channels' axis.
+            n_steps (integer, optional): Number of guessed steps, if None will match seq_len. Defaults to None.
+
+        Returns:
+            HgnResult: Bundle of the intermediate and final results of the HGN output.
+        """
+        rollout_batch = concat_rgb(rollout_batch)
```
fedetask, line 68: This should be removed once the environments output the correct shape, right?

train.py
```diff
@@ -0,0 +1,125 @@
...
+from environments.environment_factory import EnvFactory
+from hamiltonian_generative_network import HGN
+from networks.inference_net import EncoderNet, TransformerNet
+from networks.hamiltonian_net import HamiltonianNet
+from networks.decoder_net import DecoderNet
+import utilities
+
+params_file = "experiment_params/default.yaml"
+
+if __name__ == "__main__":
```
fedetask: At some point we will need to have this in a function, e.g. if we want to run multiple experiments automatically

train.py
```diff
@@ -0,0 +1,125 @@
...
+    decoder = DecoderNet(
+        in_channels=params["networks"]["transformer"]["out_channels"],
+        out_channels=params["rollout"]["n_channels"],
+        **params["networks"]["decoder"])
+
+    # Define HGN integrator
+    integrator = utilities.integrator.Integrator(
+        delta_t=params["rollout"]["delta_time"],
+        method=params["integrator"]["method"])
+
```
fedetask, line 54: Maybe we'll need to have this in a function of HGN, something like a static function `HGN.from_params()` that returns an `HGN` instance with all the attributes instantiated from the yaml configuration

utilities/integrator.py
```diff
@@ -0,0 +1,118 @@
...
+        Returns:
+            tuple(torch.Tensor, torch.Tensor): Position and momentum time derivatives: dq_dt, dp_dt.
+        """
+        # Compute energy of the system
+        energy = hnn(q=q, p=p)
+
+        # dq_dt = dH/dp
+        dq_dt = torch.autograd.grad(energy, p, create_graph=True, retain_graph=True)[0]
+        # dp_dt = -dH/dq
+        dp_dt = -torch.autograd.grad(energy, q, create_graph=True, retain_graph=True)[0]
```
fedetask: Change this as we discussed for allowing minibatches
```suggestion
        dq_dt = torch.autograd.grad(energy, p, create_graph=True, retain_graph=True,
            grad_outputs=torch.ones_like(energy))[0]
        # dp_dt = -dH/dq
        dp_dt = -torch.autograd.grad(energy, q, create_graph=True, retain_graph=True,
            grad_outputs=torch.ones_like(energy))[0]
```

fedetask (changes requested):
Nice! There are a few small things to be fixed, then we'll need to move the training into a more structured function/class.

environments/environment_factory.py
OleguerCanal: os.path.dirname(os.path.dirname( 2 times is an error or intended?

hamiltonian_generative_network.py
OleguerCanal: Yes

environments/environment_factory.py
fedetask: It's intended because dirname takes the name of the directory of the file. So if `__file__ = 'Hamiltonian-Generative-Networks/networks/inference_net.py'`, then `os.path.dirname(__file__) = Hamiltonian-Generative-Networks/networks/`, and using it twice gives `os.path.dirname(os.path.dirname(__file__)) = 'Hamiltonian-Generative-Networks'` which is the root of the project.

fedetask (approved):
Everything is good now

PR #28, opened 2020-10-18
Offline data generation

charlio23:
Offline data generator

### DIFF
--- a/environments/datasets.py
+++ b/environments/datasets.py
@@ -1,3 +1,6 @@
+import os
+
+import numpy as np
 import torch
 from torch.utils.data import Dataset
 from torchvision import transforms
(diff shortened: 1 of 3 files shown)


environments/datasets.py
```diff
@@ -86,6 +90,23 @@ def __getitem__(self, i):
...
+        self.data_mean = .5
+        self.data_std = .5
+
+    def __len__(self):
+        return len(self.file_list)
+
+    def __getitem__(self, i):
+        rolls = np.load(os.path.join(
+            self.root_dir, self.file_list[i]))['arr_0']
+        rolls = (rolls - self.data_mean) / self.data_std
```
OleguerCanal, line 106: Dont normalize the data, leave it from 0 to 1

charlio23: fixed, now mean and std are set as a parameter

stathius: Agree with @OleguerCanal, normalizations should be done with  `transform.Normalize()`

PR #30, opened 2020-10-19
Standardize shapes and types, add variational training as parameter

fedetask:
The purpose of this pull request is to clean the code, standardize shapes and types, and add the option to simply switch between variational and deterministic training.

- The training can be now done in variational or deterministic mode by changing the `variational=True` parameter in `train()`. This is propagated to `HGN.fit()` and `EncoderNet.forward()`, that accept a parameter to decide whether to sample and compute the KL loss, or to just return the mean of the encoding distribution and not compute the KL loss.
- Shapes are now standardized as follows:
    - EnvironmentSampler returns tensors of shape `(batch_size, seq_len, channels, height, width)`
    - HGN takes inputs and returns output both of shape `(batch_size, seq_len, channels, height, width)`. Concatenation along channel dimension is done internally.
    - The `Environment.visualize_rollout()` method accepts a `numpy.ndarray` of shape  `(seq_len, height, width, channels)`, while the `HgnResult.visualize()` method takes care of converting a batch of rollouts into a single sequence with the correct shape for `Environment.visualize_rollout()`
- The module `conversions.py` provides functions for shape conversions
- Now `train()` takes a `dtype` parameter that determines the type of **all** tensors used in training (including those returned by the `EnvironmentSampler`)


environments/datasets.py
```diff
@@ -1,10 +1,17 @@
+import sys
+import os
```
OleguerCanal:
```suggestion
import os
import sys
```

hamiltonian_generative_network.py
```diff
@@ -95,38 +98,48 @@ def forward(self, rollout_batch, n_steps=None):
             prediction.append_reconstruction(x_reconstructed)
         return prediction
 
-    def fit(self, rollouts):
+    def fit(self, rollouts, variational=True):
         """Perform a training step with the given rollouts batch.
 
+        TODO: Move the whole fit() code inside forward if adding hooks to the training, as direct
```
OleguerCanal, line 108: I'd prefer them to be separate so I can forward test rollouts. And try to extend the evolution in time.

hamiltonian_generative_network.py
```diff
@@ -55,25 +56,27 @@ def __init__(self,
...
-            n_steps (integer, optional): Number of guessed steps, if None will match seq_len. Defaults to None.
+            rollout_batch (torch.Tensor): Minibatch of rollouts as a Tensor of shape
+                (batch_size, seq_len, channels, height, width).
+            n_steps (integer, optional): Number of guessed steps, if None will match seq_len.
+                Defaults to None.
+            variational (bool): Whether to sample from the encoder distribution or take the mean.
 
         Returns:
-            HgnResult: Bundle of the intermediate and final results of the HGN output.
+            An HgnResult object containing data of the forward pass over the given minibatch.
```
OleguerCanal:
```suggestion
            (utilities.HgnResult): An HgnResult object containing data of the forward pass over the given minibatch.
```

Why don't you like the
```
returns:
(type): Bla bla
```
format? I would use it for consistency or if there is a good reason to use it, change all the others as well.

networks/inference_net.py
```diff
@@ -117,11 +118,14 @@ def forward(self, x):
...
-        z = mean + stddev * epsilon
-        return z, mean, log_var
+        if sample:
+            log_var = self.out_logvar(x)
+            stddev = torch.exp(0.5 * log_var)
+            epsilon = torch.randn_like(mean)
+            z = mean + stddev * epsilon
+            return z, mean, log_var
+        else:
+            return mean, None, None  # Return None to ensure that they're not used in loss
```
OleguerCanal:
```suggestion
        if not sample:
            return mean, None, None  # Return None to ensure that they're not used in loss
        log_var = self.out_logvar(x)
        stddev = torch.exp(0.5 * log_var)
        epsilon = torch.randn_like(mean)
        z = mean + stddev * epsilon
        return z, mean, log_var
```

train.py
```diff
@@ -6,6 +6,7 @@
 import torch
 import tqdm
 import yaml
+import numpy as np
```
```suggestion
import numpy as np
import torch
import tqdm
import yaml
```

Alphabetical

train.py
```diff
@@ -18,32 +19,33 @@
...
-
-def train(params):
-    """Instantiate and train the HGN.
+def train(params, variational, dtype=torch.float):
+    """Instantiate and train the Hamiltonian Generative Network.
 
     Args:
-        params (dictionary): Experiment parameters (see experiment_params folder)
+        params (dict): Experiment parameters (see experiment_params folder).
+        variational (bool): Whether to perform variational or deterministic training.
```
OleguerCanal: This should not be an extra parameter. Include it in the params dict in the yaml file

train.py
```diff
@@ -18,32 +19,33 @@
...
+        dtype (torch.dtype): Data type to be used in tensor computations.
```
OleguerCanal: Same as above

train.py
```diff
@@ -18,32 +19,33 @@
...
-        **params["networks"]["transformer"]).to(device)
-    hnn = HamiltonianNet(**params["networks"]["hamiltonian"]).to(device)
+        **params["networks"]["transformer"], dtype=dtype).to(device)
+    hnn = HamiltonianNet(**params["networks"]["hamiltonian"], dtype=dtype).to(device)
     decoder = DecoderNet(
         in_channels=params["networks"]["transformer"]["out_channels"],
-        out_channels=params["rollout"]["n_channels"],
-        **params["networks"]["decoder"]).to(device)
+        out_channels=params["rollout"]["n_channels"], **params["networks"]["decoder"],
+        dtype=dtype).to(device)
```
OleguerCanal: Is this in pep8 format?

train.py
```diff
@@ -111,12 +113,15 @@ def train(params):
     # Initialize tensorboard writer
     pbar = tqdm.tqdm(data_loader)
     for i, rollout_batch in enumerate(pbar):
-        rollout_batch = rollout_batch.float().to(device)
-        error, kld, prediction = hgn.fit(rollout_batch)
+        # rollout_batch has shape (batch_len, seq_len, channels, height, width)
+        rollout_batch = rollout_batch.to(device)
+        error, kld, prediction = hgn.fit(rollout_batch, variational=variational)
         training_logger.step(losses=(error, kld),
```
OleguerCanal, line 131: Maybe passing none to this crashes the program, did you test it?

utilities/hgn_result.py
```diff
@@ -55,22 +65,18 @@ def append_reconstruction(self, reconstruction):
...
-            reconstruction (torch.Tensor): Decoder of the HGN output
+            reconstruction (torch.Tensor): Tensor of shape (seq_len, channels, height, width)
+                containing the reconstructed rollout.
         """
-        if self.reconstructed_rollout is None:
-            self.reconstructed_rollout = reconstruction
-        else:
-            self.reconstructed_rollout = torch.cat(
-                (self.reconstructed_rollout, reconstruction), dim=1)
+        assert self.reconstruction_ptr < self.reconstructed_rollout.shape[1],\
```
OleguerCanal, line 71: I dont understand why do you limit this. What if I wanna do rollouts longer than the inputed sequence to analyse how well the system continues to evolve after the inputed data? Same with going backward in time before the provided rollout.

utilities/training_logger.py
```diff
@@ -19,24 +22,26 @@ def __init__(self, hyper_params, loss_freq=100, rollout_freq=1000):
         self.rollout_freq = rollout_freq
 
     def step(self, losses, rollout_batch, prediction):
-        """Perform a logging step: Update inner iteration counter and log info if needed
+        """Perform a logging step: update inner iteration counter and log info if needed.
```
OleguerCanal, line 25: Maybe keep the caps after ":" or change all of them for consistency

utilities/training_logger.py
```diff
@@ -19,24 +22,26 @@ def __init__(self, hyper_params, loss_freq=100, rollout_freq=1000):
...
+            rollout_batch (torch.Tensor): Batch of ground truth rollouts, as a Tensor of shape
+                (batch_size, seq_len, channels, height, width).
+            prediction (utilities.hgn_result.HgnResult): The HgnResult object containing data of
+                the models forward pass on the rollout_batch.
         """
         if self.iteration % self.loss_freq == 0:
             self.writer.add_scalar('data/reconstruction_loss', losses[0],
                                    self.iteration)
-            self.writer.add_scalar('data/kld_loss', losses[1], self.iteration)
+            if losses[1] is not None:
```
OleguerCanal, line 37: Nice

OleguerCanal (changes requested):
Very good, just minor comments

hamiltonian_generative_network.py
fedetask: Yes yes, it's okay like this for now. But if we will ever use hooks we cannot directly call the `forward()` of the model I think (see [this discussion](https://discuss.pytorch.org/t/any-different-between-model-input-and-model-forward-input/3690)).

hamiltonian_generative_network.py
fedetask: That's actually not a bad idea. I had changed it because most of our code was not using it and do not often see the `Returns` section in that format. But we can use it, it's actually quite clear.

train.py
fedetask: I think it is as in the [code layout section](https://www.python.org/dev/peps/pep-0008/#id16) they say to not put arguments in the first line when they can't fit in a single line.

We should actually define what is our limit line length.

train.py
fedetask: Yep, in `TrainingLogger` I save the kl loss only if it's not None

train.py
fedetask: Yes we will have to, but I thought it would have broken all the things we have running now. Maybe we can check it together this afternoon.

utilities/hgn_result.py
fedetask: The first reason I changed the code is that using `torch.cat()` every time we append means that we are copying the whole tensor at every call.

`self.reconstructed_rollout` is initially instantiated with the shape of the batch, so it will always work as long as you use the same length of your ground truth data. I haven't thought about the issue of testing with a length higher than your ground truth data. But wouldn't be better, in that case, to just take a longer sequence as ground truth? Otherwise we cannot compare it.

utilities/training_logger.py
fedetask: Ok

train.py
OleguerCanal: I'll do it

PR #36, opened 2020-10-20
Gradient flow debug, take type and variational from parameter file

fedetask:


utilities/debug_utils.py
```diff
@@ -0,0 +1,126 @@
+import yaml
+import sys
+import os
+
+from matplotlib import pyplot as plt
+import numpy as np
+from matplotlib.lines import Line2D
+import torch
```
OleguerCanal:
```suggestion
from matplotlib import pyplot as plt
from matplotlib.lines import Line2D
import numpy as np
import torch
```

utilities/debug_utils.py
```diff
@@ -0,0 +1,126 @@
+import yaml
+import sys
+import os
```
OleguerCanal:
```suggestion
import os
import sys
import yaml
```

train.py
```diff
@@ -17,30 +17,15 @@
...
+from utilities import debug_utils
 
 
-def train(params):
+def load_hgn(params, device, dtype):
     """Instantiate and train the Hamiltonian Generative Network.
 
     Args:
         params (dict): Experiment parameters (see experiment_params folder).
     """
```
OleguerCanal, line 29: Docstring

OleguerCanal (changes requested):

PR #37, opened 2020-10-20
Add offline data generation, refactor inference_net, and fixes to chaotic pendulum

### LINKED ISSUE #22: Add offline data generation option
As we discussed, data generation might become a bottleneck

OleguerCanal:


environments/datasets.py
```diff
@@ -91,6 +93,20 @@ def __getitem__(self, i):
...
+        self.root_dir = root_dir
+        self.file_list = os.listdir(root_dir)
+
+    def __len__(self):
+        return len(self.file_list)
+
+    def __getitem__(self, i):
+        rolls = np.load(os.path.join(
+            self.root_dir, self.file_list[i]))['arr_0']
+        return rolls.transpose((0, 3, 1, 2))
```
fedetask, line 103: Could you add a brief description of what does this load? From the shape it looks like a single rollout. Are the rollouts stored separately or in minibatches? And what is `arr_0` ?

environments/generate_data.py
```diff
@@ -0,0 +1,56 @@
...
+num_train_samples = 1000
+num_test_samples = 100
+number_of_frames = 20
+delta_time = 0.125
+img_size = 32
+noise_std = 0.
+radius_bound = (.5, 1.5)
+color = False
+world_size = 1.5
+environment = Pendulum(mass=.5, length=1, g=3)
```
fedetask: On a second thought these parameters should definitely be taken from the yaml file? We risk generating the dataset with the wrong parameters. In fact, `number_of_frames` should be 30, not 20.

experiment_params/default.yaml
```diff
@@ -17,7 +17,9 @@ dataset:
   noise_std: 0
   radius_bound: [1.3, 2.3]
```
fedetask: Are these the default parameters of the paper?

tests/test_conversions.py
```diff
@@ -14,7 +14,8 @@ def test_to_channels_last():
             for h in range(height):
                 for w in range(width):
                     for c in range(channels):
-                        assert tensor[b, s, c, h, w] == converted[b, s, h, w, c]
+                        assert tensor[b, s, c, h, w] == converted[b, s, h, w,
+                                                                  c]
```
fedetask: Is autopep doing this? It doesn't look very nice

tests/test_conversions.py
```diff
@@ -27,7 +28,8 @@ def test_to_channels_first():
             for h in range(height):
                 for w in range(width):
                     for c in range(channels):
-                        assert tensor[b, s, h, w, c] == converted[b, s, c, h, w]
+                        assert tensor[b, s, h, w, c] == converted[b, s, c, h,
+                                                                  w]
```
fedetask: Same here

utilities/hgn_result.py
```diff
@@ -6,28 +8,29 @@
...
         """
         self.input = None
         self.z_mean = None
         self.z_logvar = None
         self.z_sample = None
         self.q_s = []
         self.p_s = []
-        self.reconstructed_rollout = torch.empty(batch_shape)
+        # TODO: Bugfix
+        self.reconstructed_rollout = torch.empty(batch_shape).to(device)
```
fedetask: Nice, I hadn't thought about it. What is that TODO?

train.py
```diff
@@ -98,50 +101,93 @@ def train(params):
...
-        dtype=dtype)
-    # Dataloader instance test, batch_mode disabled
-    data_loader = torch.utils.data.DataLoader(trainDS,
-                                              shuffle=False,
-                                              batch_size=None)
+    # Either generate data on-the-fly or load the data from disk
+    data_loader = None
+    if params["dataset"]["on_the_fly_data"]:
+        dataset_len = params["optimization"]["epochs"] * params[
+            "optimization"]["batch_size"]
```
fedetask: Should we just have `dataset_len` as parameter in the yaml file? It should be independent of the batch size we use. It also makes it difficult to understand how many samples will be used. For example, in `default.yaml`, `epochs=5` and `batch_size=16`, so `dataset_len=80` rollouts. But if `batch_size=32` then the training steps be doubled.

fedetask (changes requested):
There are some things that I think are important issues. Please check this before starting tonight's trainings.

train.py
```diff
@@ -98,50 +101,93 @@ def train(params):
...
+            img_size=params["dataset"]["img_size"],
+            color=params["rollout"]["n_channels"] == 3,
+            noise_std=params["dataset"]["noise_std"],
+            radius_bound=params["dataset"]["radius_bound"],
+            world_size=params["dataset"]["world_size"],
+            seed=None)
+
+        data_loader = torch.utils.data.DataLoader(trainDS,
+                                                  shuffle=False,
+                                                  batch_size=None)
```
fedetask: This is a weird behavior. DataLoader expects this to be an `int`, but if I put `batch_size=1` the returned rollouts have an additional dimension. This is because `EnvironmentSampler` should return single rollouts in `__get_item__()` but it returns batches instead. Therefore, `DataLoader` adds a dimension corresponding to the batch size. 
I don't know why it works with `batch_size=None`, but it shouldn't according to [the docs](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader). 

We should make `EnvironmentSampler` return single samples so that we can use the `batch_size` parameter of the data loader

fedetask:
Give also a look to PR #37, it's a very short proposal on how we can fix the issue of dataset length

fedetask:
Also, we should remove the configurations that we don't need

tests/test_conversions.py
OleguerCanal: yes, fixed it

environments/datasets.py
charlio23: `arr_0` is a default name for the npz keys when no name is specified when saving npz files.

`__getitem__` method loads one sample of the dataset. Therefore, we load one rollout and prepare it to forward it to the network. Later, the dataloader samples `batch_size` samples and concatenates them automatically.

environments/generate_data.py
charlio23: of course

experiment_params/default.yaml
charlio23: it depends on the environment. In the case of the pendulum it is. In the case of the spring it is (0.1, 1.0) ,etc.

experiment_params/default_online.yaml
```diff
@@ -16,14 +16,12 @@ dataset:
...
-# Define rollout
-rollout:
-  seq_length: 30
-  delta_time: 0.1
-  n_channels: 3
+  num_train_samples: 50000  # Total number of rollouts used when training on-line.
+  num_test_samples: 100  # Number of test samples.
+  rollout:
+    seq_length: 30
+    delta_time: 0.1
```
charlio23: `delta_time = 0.125`
```suggestion
    delta_time: 0.125
```

hamiltonian_generative_network.py
```diff
@@ -26,25 +26,31 @@ def __init__(self,
                  integrator,
                  optimizer,
                  loss,
+                 device,
+                 dtype,
                  seq_len,
                  channels=3):
         """Instantiate a Hamiltonian Generative Network.
```
charlio23, line 34: In line 152-153:

`        reconstruction_error_np = reconstruction_error.detach().cpu().numpy()
        kl_div_np = kl_div.detach().cpu().numpy() if kl_div is not None else None`
otherwise this blows up when using gpu.

utilities/hgn_result.py
```diff
@@ -68,6 +70,7 @@ def append_reconstruction(self, reconstruction):
             reconstruction (torch.Tensor): Tensor of shape (seq_len, channels, height, width)
                 containing the reconstructed rollout.
         """
+        # TODO Fix this error. Workaround solution
```
charlio23: Remove this comment
```suggestion

```

charlio23 (commented):
Everything looks fine for me.

train.py
fedetask: Ok this was not a problem, `DataLoader` can accept `batch_size=Nose` and in that case it won't produce the additional dimension at all.

PR #38, opened 2020-10-20
Add potential Learning option

### LINKED ISSUE #33: Try hamiltonian net to only take q and adapt leapfrog integrator
Instead of learning H(p, q). If we assume H is separable as H(p, q) = V(q) + T(p), we only need to learn V for the leapfrog integrator.
Lets see if this works better

fedetask:
We called Potential Learning the technique in which the Hamiltonian network only takes q as parameter. The learned value should then represent only the potential energy. We then use Leapfrog integration to perform the update steps on q and p. Note that Leapfrog integration does not require dH/dp and thus Potential Learning looks reasonable to us.


experiment_params/default_environments/three_bodies.yaml
```diff
@@ -0,0 +1,5 @@
+environment:
+  name: "NObjectGravity"
+  mass: [0.5, 0.5, 0.5]
```
charlio23: I assume that this is fixed in master because of the offline data PR

networks/hamiltonian_net.py
```diff
@@ -105,9 +109,14 @@ def forward(self, q, p):
         Returns:
             A (batch_size, 1) shaped tensor with the energy for each input in the batch.
         """
-        x = torch.cat(
-            (q, p),
-            dim=1)  # Concatenate q and p to obtain a N x 2C x H x W tensor
+        if p is None and not self.potential_learning:
+            raise ValueError('Receiving p = None when the network is in potential learning mode.')
```
charlio23, line 113:
```suggestion
            raise ValueError('Receiving p = None when the network is not in potential learning mode.')
```

train.py
```diff
@@ -139,6 +144,7 @@ def train(params, cpu=False, resume=False):
     # Read parameters
     with open(params_file, 'r') as f:
         params = yaml.load(f, Loader=yaml.FullLoader)
+    params['potential_learning'] = args.potential_learning
```
charlio23: The yaml file already contrains the potential learning parameter inside the hamiltonian configuration.

`params['networks']['hamiltonian']['potential_learning']`

Moreover, are you sure that this parameter is passed when initializing the hgn? I saw changes in the constructor for the hamiltonian network but not in `loader.py` (the script that instantiates the hamiltonian net).

charlio23 (commented):
Nice work!

experiment_params/default_offline.yaml
```diff
@@ -37,6 +37,7 @@ networks:
     n_filters: [32, 64, 64, 64, 64, 64]  # first + hidden
     kernel_sizes: [3, 3, 3, 3, 3, 3]  # first + hidden + last
     strides: [1, 1, 1, 1, 1, 1]  # first + hidden + last
+    potential_learning: True
```
OleguerCanal:
```suggestion
    potential_learning: True  # If true integrator must be Leapfrog
```

experiment_params/default_online.yaml
```diff
@@ -49,6 +47,7 @@ networks:
     n_filters: [32, 64, 64, 64, 64, 64]  # first + hidden
     kernel_sizes: [3, 3, 3, 3, 3, 3]  # first + hidden + last
     strides: [1, 1, 1, 1, 1, 1]  # first + hidden + last
+    potential_learning: False
```
OleguerCanal, line 50:
```suggestion
    potential_learning: False  # If True integrator must be Leapfrog
```

train.py
```diff
@@ -43,13 +43,14 @@ def train(params, cpu=False, resume=False):
...
-    if "environment" in params:
-        print("Training with ONLINE data...")
-        train_data_loader, test_data_loader = get_online_dataloaders(params)
-    else:
+    if "train_data" in params["dataset"]:
         print("Training with OFFLINE data...")
         train_data_loader, test_data_loader = get_offline_dataloaders(params)
-
+    else:
+        assert "environment" in params, "Nor environment nor train_data are specified in the " \
```
```suggestion
        assert "environment" in params, "Neither environment nor train_data are specified in the " \
```
jeje

train.py
```diff
@@ -130,17 +131,29 @@ def train(params, cpu=False, resume=False):
...
     params_file = args.params[0] if args.params is not None else DEFAULT_PARAM_FILE
     # Read parameters
     with open(params_file, 'r') as f:
         params = yaml.load(f, Loader=yaml.FullLoader)
 
     if args.name is not None:
         params['experiment_id'] = args.name[0]
+    if args.epochs is not None:
+        params['optimization']['epochs'] = args.epochs[0]
+    params['networks']['hamiltonian']['potential_learning'] = args.potential_learning
```
OleguerCanal: Wont this always overwrite params['networks']['hamiltonian']['potential_learning'] ?

train.py
```diff
@@ -130,17 +131,29 @@ def train(params, cpu=False, resume=False):
...
+    if args.resume is not None:
+        raise NotImplementedError('Resume training from command line is not implemented yet')
+
     # Train HGN network
     hgn = train(params, cpu=args.cpu)
```
OleguerCanal: Same here, I think it would be better to also be able to define the device in the params.yaml.

Maybe we can have a param like:
```yaml
device: 'cpu'
```

or

```yaml
device: 'cuda:0'
```

Instead of the cuda_id. WHat do you think?

utilities/integrator.py
```diff
@@ -6,7 +6,7 @@ class Integrator:
     """
     METHODS = ["Euler", "RK4", "Leapfrog"]
 
-    def __init__(self, delta_t, method="Euler"):
+    def __init__(self, delta_t, method="Leapfrog"):
```
OleguerCanal, line 9: Nice haha

utilities/integrator.py
```diff
@@ -31,20 +31,24 @@ def _get_grads(self, q, p, hnn):
...
         # Compute energy of the system
-        energy = hnn(q=q, p=p)
-
-        # dq_dt = dH/dp
-        dq_dt = torch.autograd.grad(energy,
-                                    p,
-                                    create_graph=True,
-                                    retain_graph=True,
-                                    grad_outputs=torch.ones_like(energy))[0]
+        energy = hnn(p=p, q=q) if not hnn.potential_learning else hnn(q=q)
```
OleguerCanal, line 43:
```suggestion
        energy = hnn(q=q, p=p) if not hnn.potential_learning else hnn(q=q, p=None)
```

OleguerCanal: Although I'm not sure it should be this guy responsibility to know how to call hnn(). If hnn is in potential learning, it should ignore p automatically.

OleguerCanal: On a second thought maybe this is less error prone since the integrator checks that leapfrog and potential_learning match. Otherwise we would need to do it somewhere else. What do you think?

OleguerCanal (changes requested):
Very nice!

Maybe lets try not merge two pending PR branches into one since basically this was two PRs at once. I think its easier to review if we keep things independent.

PR #42, opened 2020-10-21
GECO

### LINKED ISSUE #21: Implement GECO to stabilize learning
Danilo J Rezende and Fabio Viola. Generalized elbo with constrained optimization, geco. Workshop on Bayesian Deep Learning, NeurIPS, 2018.

stathius:
Implementation of the GECO algorithm from the paper Rezende, D. J., & Viola, F. (2018). Taming VAEs. Retrieved from http://arxiv.org/abs/1810.00597


hamiltonian_generative_network.py
```diff
@@ -125,7 +150,7 @@ def fit(self, rollouts, variational=True):
         prediction = self.forward(rollout_batch=rollouts, variational=variational)
         
         # Compute frame reconstruction error
-        reconstruction_error = self.loss(input=prediction.input,
+        reconstruction_error = self.reconstruction_loss(input=prediction.input,
```
OleguerCanal: input is not a keyword argument of reconstruction_loss

utilities/losses.py
```diff
@@ -0,0 +1,37 @@
...
+        prediction (Tensor) The prediction of the model
+
+    Returns:
+        (Tensor): MSE loss
+    """
+    # Compute the sum error across each channel in the batch and then average
+    # last_dim = len(target.shape) - 1
+    # loss_per_channel = torch.sum(torch.pow(prediction - target, 2), [last_dim - 1, last_dim])
+    # return torch.mean(loss_per_channel)
+    return torch.mean(torch.pow(prediction - target, 2))
```
OleguerCanal, line 17: Why not use torch.nn.MSELoss()?

utilities/losses.py
```diff
@@ -0,0 +1,37 @@
...
+        mu (torch.Tensor): the part of the latent vector that corresponds to the mean
+        logvar (torch.Tensor): the log of the variance (sigma squared)
+
+    Returns:
+        (torch.Tensor): KL divergence.
+    """
+    return torch.mean(-0.5 * torch.sum(1 + logvar - mu.pow(2) - logvar.exp(), 1))
+
+def geco_constraint(target, prediction, tol):
+    """Computes the constraint for the geco algorithm
```
OleguerCanal, line 34: Args and Returns missing in docstring

utilities/training_logger.py
```diff
@@ -40,11 +40,9 @@ def step(self, losses, rollout_batch, prediction, model):
                 the models forward pass on the rollout_batch.
         """
         if self.iteration % self.loss_freq == 0:
-            self.writer.add_scalar('data/reconstruction_loss', losses[0],
-                                   self.iteration)
-            if losses[1] is not None:
-                self.writer.add_scalar('data/kld_loss', losses[1],
-                                       self.iteration)
+            for loss_name, loss_value in losses.items():
```
OleguerCanal, line 43: Nice!

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
-    # Initialize tensorboard writer
-    model_save_file = os.path.join(params["model_save_dir"],
-                                   params["experiment_id"])
-
-    # TRAIN
-    for ep in range(params["optimization"]["epochs"]):
-        print("Epoch %s / %s" %
-              (str(ep + 1), str(params["optimization"]["epochs"])))
-        pbar = tqdm.tqdm(train_data_loader)
+class hgn_trainer:
```
```suggestion
class HgnTrainer:
```

train.py
```diff
@@ -143,4 +237,5 @@ def train(params, cpu=False, resume=False):
     if args.name is not None:
         params['experiment_id'] = args.name[0]
     # Train HGN network
-    hgn = train(params, cpu=args.cpu)
+    trainer = hgn_trainer(params, cpu=args.cpu)
```
```suggestion
    trainer = HgnTrainer(params, cpu=args.cpu)
```

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+                'params': self.hgn.decoder.parameters(),
+                'lr': params["optimization"]["decoder_lr"]
+            },
+        ]
+        self.optimizer = torch.optim.Adam(optim_params)
+
+    def training_step(self, rollouts):
+        """Perform a training step with the given rollouts batch.
+
+        TODO: Move the whole fit() code inside forward if adding hooks to the training, as direct
```
OleguerCanal, line 88: Does this todo still apply?

experiment_params/default_online.yaml
```diff
@@ -67,3 +66,8 @@ optimization:
   transformer_lr: 1.5e-4
   hnn_lr: 1.5e-4
   decoder_lr: 1.5e-4
+
+geco:
+  alpha: 0.99
```
OleguerCanal, line 71: Can you add a small description of each parameter?

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+                'lr': params["optimization"]["hnn_lr"]
+            },
+            {
+                'params': self.hgn.decoder.parameters(),
+                'lr': params["optimization"]["decoder_lr"]
+            },
+        ]
+        self.optimizer = torch.optim.Adam(optim_params)
+
+    def training_step(self, rollouts):
```
OleguerCanal, line 85: maybe this should be a protected method

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+        if "environment" in self.params:
+            print("Training with ONLINE data...")
+            self.train_data_loader, self.test_data_loader = get_online_dataloaders(self.params)
+        else:
+            print("Training with OFFLINE data...")
+            self.train_data_loader, self.test_data_loader = get_offline_dataloaders(self.params)
+
+        # Initialize training logger
+        self.training_logger = TrainingLogger(hyper_params=self.params,
+                                         loss_freq=1,
```
OleguerCanal, line 57: Did you autopep?

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+            rec_loss = reconstruction_loss(target=prediction.input, 
+                                       prediction=prediction.reconstructed_rollout)
+            losses = {'loss/train' : rec_loss}
+
+        return losses, hgn_output
+
+    def fit(self):
+        """
+        The trainer fits an HGN.
+        """
```
OleguerCanal, line 150:
```suggestion
    def fit(self):
        """The trainer fits an HGN.
        """
```

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+            A dictionary of losses and the model's prediction of the rollout. The reconstruction loss and
+            KL divergence are floats and prediction is the HGNResult object with data of the forward pass.
+        """
+
+        hgn_output = self.hgn.forward(rollout_batch=rollouts)
+        target = hgn_output.input
+        prediction = hgn_output.reconstructed_rollout
+        
+        if self.params["networks"]["variational"]:    
+            C, rec_loss = geco_constraint(target, prediction, self.params["geco"]["tol"])
```
OleguerCanal, line 107: Please, try to put more meaningful names to variables. `C` alone is hard to know what it is

train.py
OleguerCanal: Also, do we need to a loss freq of 1?

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+    
+            # Compute moving average of constraint C
+            if self.C_ma is None:
+                self.C_ma = C
+            else:
+                # not exactly sure if i should detach here on in the expression below 
+                self.C_ma = self.params["geco"]["alpha"] * self.C_ma.detach() + \
+                            (1 - self.params["geco"]["alpha"]) * C
+
+            C = C + (self.C_ma - C)
```
OleguerCanal, line 122: Isnt that the same as:
```python
C = self.C_ma
```
Is there some value in adding the extra computation?

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+            
+            # Compute losses
+            train_loss = kld + self.langrange_multiplier * C
```
OleguerCanal, line 125: Probably im missing something, but I dont really see the point of updating C if you are just gonna use it here.
```suggestion
            train_loss = kld + self.langrange_multiplier * self.C_ma
```

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+            logvar = hgn_output.z_logvar
+            kld = kld_loss(mu, logvar)
+    
+            # Compute moving average of constraint C
+            if self.C_ma is None:
+                self.C_ma = C
+            else:
+                # not exactly sure if i should detach here on in the expression below 
+                self.C_ma = self.params["geco"]["alpha"] * self.C_ma.detach() + \
+                            (1 - self.params["geco"]["alpha"]) * C
```
OleguerCanal, line 120: is `self.C_ma` supposed to be detached? maybe you should detach C first and then use it to update C_ma so C_ma is detached

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+                      'loss/C_cur': C_curr,
+                      'loss/C_ma': self.C_ma,
+                      'loss/rec': rec_loss,
+                      'other/langrange_mult': self.langrange_multiplier
+                     }
+
+            # clamping the langrange multiplier to avoid inf values
+            self.langrange_multiplier = torch.clamp(self.langrange_multiplier * 
+                                                    torch.exp(self.params["geco"]["langrange_multiplier_param"]
+                                                    * self.C_ma.detach()), 1e-10, 1e10)
```
OleguerCanal, line 138: Same as before, I think you want C_ma detached from the beginning. You can use C.detach() to update it

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+            KL divergence are floats and prediction is the HGNResult object with data of the forward pass.
+        """
+
+        hgn_output = self.hgn.forward(rollout_batch=rollouts)
+        target = hgn_output.input
+        prediction = hgn_output.reconstructed_rollout
+        
+        if self.params["networks"]["variational"]:    
+            C, rec_loss = geco_constraint(target, prediction, self.params["geco"]["tol"])
+            C_curr = C.item()
```
OleguerCanal, line 108: Try to define variables closer to where they are used

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+                self.optimizer.step()
+
+                # Log progress
+                self.training_logger.step(losses=losses,
+                                     rollout_batch=rollout_batch,
+                                     prediction=prediction,
+                                     model=self.hgn)
+
+                # Progress-bar msg
+                msg = ", ".join([f"{k}: {v:.2e}" for k,v in losses.items() if v is not None])
```
OleguerCanal, line 179: Nice!

train.py
```diff
@@ -22,88 +22,182 @@ def _avoid_overwriting(experiment_id):
...
+            # Save model
+            self.hgn.save(model_save_file)
+
+        self.test()
+        return self.hgn
+
+    def test(self):
+        """
+        Test after the training is finished.
+        """
```
```suggestion
        """Test after the training is finished.
        """
```

OleguerCanal (changes requested):
Very nice changes! Just some minor comments of things I did not fully understand of the geco implementation. If you can try to add more comments on the code.

utilities/losses.py
stathius: Well, in the beginning i wasn't going for a pixel wise loss but a channel wise. So it's kind of "legacy".

train.py
stathius: Agree. I was thinking to rename that but then forgot.

train.py
stathius: Which var are you referring to?

OleguerCanal: C_curr

utilities/losses.py
OleguerCanal: So, should we just use orch.nn.MSELoss()?

stathius: I guess we could for readability, I will change it.

train.py
stathius: I am not sure what you suggest can you provide a snippet. Have in mind that we *do* want gradients to backpropagate through C and C_ma but not through the previous values of C_ma. I am not 100% sure but this is my best guess for what they are trying to achieve.  I've been playing with some other ways to do this and this seems to be giving good behavior and results on MNIST. In any case I'm not sure if detaching C completely would be correct.

train.py
stathius: See prev answer.

train.py
stathius: It is used to keep the value of C before changing it. I could put it before changing C if that is better. It is mostly for logging.

PR #44, opened 2020-10-22
Generate data from command line

fedetask:
`generate_data.py` can be fully used with command line arguments. See the `README.txt` for a description.


generate_data.py
```diff
@@ -0,0 +1,170 @@
...
+        env_params = _read_params(DEFAULT_ENVIRONMENTS_PATH + args.env[0] + '.yaml')
+        params['environment'] = env_params['environment']
+    if args.env_spec is not None:
+        for spec in args.env_spec:
+            key, value = spec.split(':')
+            params['environment'][key] = ast.literal_eval(value)
+
+
+if __name__ == '__main__':
+    DATASETS_ROOT = os.path.join(os.path.dirname(__file__), 'datasets')
```
charlio23, line 74: Could we make this be a parameter in the yaml file?

In my case i need to load the datasets in a different location than "datasets". This change also applies to offline dataset loader called in the train.py file. I might open a new PR for that.

experiment_params/default_environments/three_bodies.yaml
```diff
@@ -0,0 +1,5 @@
+environment:
+  name: "NObjectGravity"
+  mass: [0.5, 0.5, 0.5]
```
charlio23:
```suggestion
  mass: [1., 1., 1.]
```

experiment_params/default_environments/two_bodies.yaml
```diff
@@ -0,0 +1,5 @@
+environment:
+  name: "NObjectGravity"
+  mass: [0.5, 0.5]
```
charlio23:
```suggestion
  mass: [1., 1.]
```

charlio23 (commented):
Nice implementation!

experiment_params/default_environments/three_bodies.yaml
fedetask: Thanks, I was not sure about the default parameters. Are all the others okay?

generate_data.py
fedetask: Can I make it a command line argument? Because it would be weird to have it the `default_online.yaml`

charlio23: oh that would be perfect.

OleguerCanal (commented):
generate_data.py is missing docstrings but I think for now its ok

charlio23 (approved):
Nice!

PR #45, opened 2020-10-22
Add 'auto' option for radius_bound

fedetask:
The `Environment.sample_random_rollouts()` now accepts `radius_bound='auto'`, that makes it use the default `radius_bound` as returned by each environment new method `get_default_radius_bound()`.

The `Environment` class now defines an abstract `get_default_radius_bound()` method that all environments must implement.

### DIFF
--- a/environments/chaotic_pendulum.py
+++ b/environments/chaotic_pendulum.py
@@ -62,7 +62,7 @@ class ChaoticPendulum(Environment):
         """Returns:
             radius_bounds (tuple): (min, max) radius bounds for the environment.
         """
-        return (1.3, 2.3)
+        return (0.5, 1.3)
 
     def _dynamics(self, t, states):
         """Defines system dynamics
--- a/environments/chaotic_pendulum.py
+++ b/environments/chaotic_pendulum.py
@@ -60,7 +60,7 @@ class ChaoticPendulum(Environment):
 
     def get_default_radius_bounds(self):
         """Returns:
-        radius_bounds (tuple): (min, max) radius bounds for the environment.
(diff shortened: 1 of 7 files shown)


environments/gravity.py
```diff
@@ -72,6 +72,17 @@ def get_max_noise_std(self):
...
+    def get_default_radius_bounds(self):
+        """Returns:
+            radius_bounds (tuple): (min, max) radius bounds for the environment.
+        """
+        if self.n_objects == 2:
+            return (0.5, 1.5)
+        elif self.n_objects == 3:
+            return (0.9, 1.2)
+        else:
+            return (1., 1.)  # TODO: Should we raise an error?
```
OleguerCanal: I think its ok, maybe print a warning?

OleguerCanal (approved):

environments/chaotic_pendulum.py
```diff
@@ -58,6 +58,12 @@ def get_max_noise_std(self):
         """Return maximum noise std that keeps the environment stable."""
         return 0.05
 
+    def get_default_radius_bounds(self):
+        """Returns:
+            radius_bounds (tuple): (min, max) radius bounds for the environment.
+        """
+        return (1.3, 2.3)
```
charlio23: for chaotic pendulum i use (0.5, 1.3). I think it generates diverse data and (1.3, 2.3) might contain too much momentum. 
```suggestion
        return (0.5, 1.3)
```

environments/gravity.py
fedetask: Ok you're right

charlio23 (approved):

PR #46, opened 2020-10-23
Hgn sampler

OleguerCanal:
Added functionality to sample rollouts from a trained HGN


hamiltonian_generative_network.py
```diff
@@ -171,14 +182,62 @@ def save(self, directory):
...
+            img_shape (tuple(int, int), optional): Size of the images, should match the trained ones. Defaults to (32, 32).
+
+        Returns:
+            (utilities.HgnResult): An HgnResult object containing data of the forward pass over the
+                given minibatch.
+        """
+        # Sample from a normal distribution the latent representation of the rollout
+        latent_shape = (1, self.encoder.out_mean.out_channels, img_shape[0],
+                        img_shape[1])
+        latent_representation = torch.randn_like(torch.empty(latent_shape))
```
fedetask:
```suggestion
        latent_representation = torch.randn(latent_shape)
```

test.py
```diff
@@ -0,0 +1,16 @@
...
+    model_to_load = "saved_models/two_bodies_default"
+    
+    integrator = Integrator(delta_t=0.125, method="Leapfrog")
+    hgn = HGN(integrator=integrator)  # If going to load, no need to specify networks
+    hgn.load(model_to_load)
+    
+    # Sample a rollout of n_steps
+    prediction = hgn.get_random_sample(n_steps=50, img_shape=(32, 32))
+    prediction.visualize()
+
```
fedetask: Should this file be in root? If yes, can you rename to something like `sample_rollouts.py`?

fedetask (changes requested):
Good job! Only two minor changes before merge

test.py
OleguerCanal: ok, good point

PR #55, opened 2020-10-23
image generation with ball smoothing rather than exponential propagation

charlio23:
Balls are now smoothed circles rather than a propagated exponential.

### DIFF
--- a/environments/chaotic_pendulum.py
+++ b/environments/chaotic_pendulum.py
@@ -1,3 +1,4 @@
+import cv2
 import numpy as np
 
 from environment import Environment, visualize_rollout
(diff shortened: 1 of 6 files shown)


environments/chaotic_pendulum.py
```diff
@@ -1,3 +1,4 @@
+import cv2
```
OleguerCanal, line 1: Add cv2 to requirements.txt

environments/environment.py
```diff
@@ -76,6 +76,13 @@ def get_default_radius_bounds(self):
         """
         raise NotImplementedError
 
+    def _world_to_pixels(self, x, y, res):
```
OleguerCanal, line 79: Nice function! Add docstring.

OleguerCanal (changes requested):
Very nice improvement!

environments/chaotic_pendulum.py
charlio23: sorryy

environments/environment.py
charlio23: :)

environments/environment.py
```diff
@@ -77,7 +77,16 @@ def get_default_radius_bounds(self):
...
     def _world_to_pixels(self, x, y, res):
+        """Maps coordinates from world space to pixel space
 
+        Args:
+            x (float): x coordinate of the world space.
+            y (float): y coordinate of the world space.
+            res (int): Image resolution in pixels (images are square).
+
+        Returns:
+            (float, float): Tuple of coordinates in pixel space.
```
OleguerCanal:
```suggestion
            (int, int): Tuple of coordinates in pixel space.
```

PR #56, opened 2020-10-24
Make geco branch compatible with master

fedetask:
This is a proposal for resolution of conflicts between `master` and `geco` that arose since we continued developing on `master`.


fedetask:
Wrong target branch

PR #62, opened 2020-10-26
Documentation

stathius:
Some documentation/comments/docstrings that were missing.

### DIFF
--- a/utilities/losses.py
+++ b/utilities/losses.py
@@ -34,12 +34,12 @@ def geco_constraint(target, prediction, tol):
     """Computes the constraint for the geco algorithm.
 
     Args:
-        target(Tensor): the rollout target
-        prediction(Tensor): the prediction of the model
-        tol(float): the tolerance we accept between target and prediction
+        target (toch.Tensor): the rollout target.
+        prediction (torch.Tensor): the prediction of the model.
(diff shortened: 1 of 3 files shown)


utilities/losses.py
```diff
@@ -31,7 +31,15 @@ def kld_loss(mu, logvar):
...
-    """Computes the constraint for the geco algorithm
+    """Computes the constraint for the geco algorithm.
+
+    Args:
+        target(Tensor): the rollout target
+        prediction(Tensor): the prediction of the model
+        tol(float): the tolerance we accept between target and prediction
+
+    Returns:
+        (Tensor): the constraing value as MSE minus the tolerance
```
OleguerCanal:
```suggestion
    Args:
        target (toch.Tensor): the rollout target.
        prediction (torch.Tensor): the prediction of the model.
        tol (float): the tolerance we accept between target and prediction.

    Returns:
        (tuple(torch.Tensor, torch.Tensor)): the constraing value as MSE minus the tolerance, and MSE.
```

OleguerCanal (changes requested):

fedetask (approved):
Great! I think we can merge this

OleguerCanal (approved):

PR #74, opened 2020-11-07
Fix model io

### LINKED ISSUE #70: Training input/output
"PS: Another small detail that might matter, is that during training, assume we use for inference 5 images and unroll for 30, than we would feed the model a sequence of length 34, where the first 5 are used for inference and the last 30 are the targets (after unrolling for 29 steps). E.g. the inference network infers the state of the system at the last image it sees (which would be the 5th in this case). "

Data rollout: 34 frames
Model input: First 5 frames
Loss for: Last 30

OleguerCanal:
Read linked issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Re-implementation of Hamiltonian Generative Networks [paper](https://arxiv.org/a
 The [train.py](train.py) script takes care of performing the training.
 To start a training, run 
 ```commandline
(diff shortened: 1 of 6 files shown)


charlio23 (approved):
Good job!

OleguerCanal:
Thanks :)

fedetask (approved):
Everything looks good!

PR #75, opened 2020-11-07
move get random sample latent represent input to device

charlio23:
simple bugfix for sampling latent representation when using gpu


OleguerCanal:
Nice!

OleguerCanal (approved):

PR #76, opened 2020-11-07
Fix hamiltonian network

fedetask:


experiment_params/train_config_default.yaml
```diff
@@ -20,11 +20,12 @@ networks:
...
   hamiltonian:
-    hidden_conv_layers: 4
+    hidden_conv_layers: 3
     in_shape: [16, 4, 4]  # Should be coherent with transformer output
-    n_filters: [32, 64, 64, 64, 64, 64]  # first + hidden
-    kernel_sizes: [3, 3, 3, 3, 3, 3]  # first + hidden + last
-    strides: [1, 1, 1, 1, 1, 1]  # first + hidden + last
+    n_filters: [32, 64, 64, 64]  # first + hidden
+    kernel_sizes: [3, 2, 2, 2, 2]  # first + hidden + last
+    strides: [1, 2, 1, 1, 1]  # first + hidden + last
```
charlio23, line 27:
```suggestion
    strides: [1, 2, 1, 1, 2]  # first + hidden + last
```
last one with stride 2

experiment_params/train_config_default.yaml
```diff
@@ -20,11 +20,12 @@ networks:
...
+    paddings: [1, 0, [0, 1, 0, 1], [0, 1, 0, 1], 0]  # first + hidden + last
```
charlio23, line 28: why is the 3rd and 4th element [0, 1, 0, 1] ?

networks/hamiltonian_net.py
```diff
@@ -44,52 +46,49 @@ def __init__(self,
...
                    'network (' + str(hidden_conv_layers + 2) + ' in this case).'
         else:
             raise ValueError(
                 'Args hidden_conv_layers, n_filters, kernel_sizes, and strides'
-                'can only be either all None, or all defined by the user.')
+                'can only be either all None, or all defined by the user.'
+            )
+        self.paddings = paddings
+        conv_paddings = [0 if isinstance(p, list) else p for p in paddings]
+        print(conv_paddings)
```
charlio23:
```suggestion
```
remove print

charlio23 (commented):
nice job

experiment_params/train_config_default.yaml
charlio23: i understand now

experiment_params/train_config_default.yaml
OleguerCanal: I think stride 2 in the last one doesn't make a lot of sense since the input is 2x2 and the kernel is 2x2

networks/hamiltonian_net.py
```diff
@@ -5,6 +5,7 @@
 
 import torch
 from torch import nn
+from torch.nn import functional as F
```
OleguerCanal, line 8:
```suggestion
from torch.nn import functional
```

OleguerCanal: F is not very informative, since we are only gonna  use this in a couple of places I would keep the functional name or something more representative than F

networks/hamiltonian_net.py
```diff
@@ -108,12 +107,15 @@ def forward(self, q, p):
         x = torch.cat(
             (q, p),
             dim=1)  # Concatenate q and p to obtain a N x 2C x H x W tensor
+        if isinstance(self.paddings[0], list):
+            x = F.pad(x, self.paddings[0])
```
OleguerCanal, line 110:
```suggestion
            x = functional.pad(x, self.paddings[0])
```

networks/hamiltonian_net.py
```diff
@@ -108,12 +107,15 @@ def forward(self, q, p):
...
         x = self.activation(self.in_conv(x))
-        for layer in self.hidden_layers:
+        for i, layer in enumerate(self.hidden_layers):
+            if isinstance(self.paddings[i + 1], list):
+                x = F.pad(x, self.paddings[i + 1])
```
OleguerCanal, line 114:
```suggestion
                x = functional.pad(x, self.paddings[i + 1])
```

networks/hamiltonian_net.py
```diff
@@ -108,12 +107,15 @@ def forward(self, q, p):
         x = torch.cat(
             (q, p),
             dim=1)  # Concatenate q and p to obtain a N x 2C x H x W tensor
+        if isinstance(self.paddings[0], list):
```
OleguerCanal, line 109: Why is the first one a special case?

OleguerCanal: Is it because its the input layer not defined by the user?

OleguerCanal (changes requested):

networks/hamiltonian_net.py
fedetask: Ok that works for me. I used `F` as it looks kind of the standard in pytorch

networks/hamiltonian_net.py
fedetask: Because `paddings` is a list that can contain a `int` value, in which case the padding is intended symmetric, or a list that specifies the padding in all 4 sides. In the first case we already take care of it in the constructor, passing the int value to the `Conv2d` constructor. If the padding is a list instead, we need to create a padding layer in the `forward()`

networks/hamiltonian_net.py
stathius: F is used widely in the community, it's less ambiguous than `functional` imo

PR #77, opened 2020-11-07
changed color rendering and smoothing

### LINKED ISSUE #69: Datasets closer to the real ones
Use the dataset example sent by the authors to make ours more similar.

charlio23:
New background and ball painting colors; matching the ones in the original implementation. Image smoothing also as close as possible to the original data.

### DIFF
--- a/environments/chaotic_pendulum.py
+++ b/environments/chaotic_pendulum.py
@@ -66,6 +66,12 @@ class ChaoticPendulum(Environment):
         """
         return (0.5, 1.3)
 
+    def _get_default_ball_color(self):
+        """Returns:
+            color_ball [(tuple)]: (R, G, B) default color balls for rendering.
+        """
+        return [(173./255, 146./255, 0.), (173./255, 0., 0.)]
+
     def _dynamics(self, t, states):
         """Defines system dynamics
 
@@ -123,6 +129,8 @@ class ChaoticPendulum(Environment):
         q = self._rollout.reshape(2, 2, -1)[0, :, :]
         length = q.shape[-1]
         vid = np.zeros((length, res, res, 3), dtype='float')
+        background_color = self._get_default_background_color()
+        ball_colors = self._get_default_ball_color()
         space_res = 2.*self.get_world_size()/res
         for t in range(length):
             coords_1 = self._world_to_pixels(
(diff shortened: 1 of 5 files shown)


environments/chaotic_pendulum.py
```diff
@@ -66,6 +66,12 @@ def get_default_radius_bounds(self):
         """
         return (0.5, 1.3)
 
+    def _get_default_ball_color(self):
```
OleguerCanal: Feels like colors could be a protected member of the abstract class, then each child uses as many colors as it needs

environments/environment.py
```diff
@@ -77,6 +77,38 @@ def get_default_radius_bounds(self):
...
+    def _get_default_ball_color(self):
+        """Returns:
+            color_ball (tuple): (R, G, B) default color ball for rendering.
+
+        Raises:
+            NotImplementedError: Class instantiation has no implementation
+        """
+        raise NotImplementedError
+
+    def _get_default_background_color(self):
```
OleguerCanal: I think its better if backround_color is a protected class member, instead of a function that returns a hard-coded value

environments/gravity.py
```diff
@@ -139,28 +145,35 @@ def _draw(self, res=32, color=True):
...
                                         self._world_to_pixels(
                                             q[n, 0, t], q[n, 1, t], res),
-                                        int(self.mass[n]*.55/space_res), (0., 1., 0.), -1)
-            vid[t] = cv2.blur(vid[t], (3, 3))
+                                        int(self.mass[n]*factor/space_res), ball_colors[2], -1)
+            if n_objects == 2:
+                vid[t] = cv2.blur(cv2.blur(vid[t], (2, 2)), (2, 2))
         if color:
-            vid += 80./255.
+            vid[:] += background_color
```
OleguerCanal: If you add the background_color to all values, don't you modify the pre-set ball colors?

OleguerCanal (changes requested):

environments/chaotic_pendulum.py
charlio23: agreed

environments/environment.py
charlio23: you mean self._background_color right?

environments/gravity.py
charlio23: Yep, the real ball color variable is the addition of the background and the ball color.

charlio23: I will change the addition of the background to be in the initialization of the vid array

charlio23: well, i have been thinking and it's better to add the background afterwards. if you think of it the B channel is always constant. Therefore, when you add an element you need to specify the B channel (which will be the B component of the background). You can set the B component of the elements to 0 if you add the background in the end.

OleguerCanal (approved):
Nice!

environments/gravity.py
OleguerCanal: Ok, looks good! Thanks for the clarification :)

environments/environment.py
OleguerCanal: yes

PR #82, opened 2020-12-12
train test loss display after training

### LINKED ISSUE #80: Display train and test losses (reconstruction and KL) when training is done
Print mean +- 95% CI

Print to tensorboard

charlio23:


utilities/statistics.py
```diff
@@ -0,0 +1,17 @@
...
+
+def mean_confidence_interval(data, confidence=0.95):
+    """Calculates mean and confidence interval from samples such that they lie within m +/- h 
+    with the given confidence.
+
+    Args:
+        data (np.array): Sample to calculate the confidence interval.
+        confidence (float): Confidence of the interval (betwen 0 and 1).
+    """
+    a = 1.0 * np.array(data)
```
OleguerCanal: Why 1.0 * np.array?

Do you wanna set the dtype??

OleguerCanal (changes requested):
Seems that test code is repeated twice, would be nice to have a single function

OleguerCanal (approved):

PR #83, opened 2021-01-19
HGN working without transformer

fedetask:
This pull request makes HGN work without the transformer. If `transformer` is not present under `networks` in the configuration, HGN will not use it. The latent encoding will be split into `q` and `p` and used. Therefore, the encoder and hamiltonian networks should handle dimensions appropriately.

I updated the `train_config_no_transformer.yaml` by making the latent encoding of 32 channels instead of 48, so that q and p are of 16 channels each.

### DIFF
--- a/hamiltonian_generative_network.py
+++ b/hamiltonian_generative_network.py
@@ -85,8 +85,10 @@ class HGN:
         prediction.set_z(z_sample=z, z_mean=z_mean, z_logvar=z_logvar)
 
         # Initial state
-        #q, p = self.transformer(z)
-        q, p = torch.split(z, z.size(1)//2, dim=1)
+        if self.transformer is not None:
+            q, p = self.transformer(z)
+        else:
+            q, p = torch.split(z, z.size(1)//2, dim=1)
         prediction.append_state(q=q, p=p)
 
         # Initial state reconstruction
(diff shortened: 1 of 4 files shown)


hamiltonian_generative_network.py
```diff
@@ -163,9 +168,12 @@ def get_random_sample(self, n_steps, img_shape=(32, 32)):
...
         # Sample from a normal distribution the latent representation of the rollout
-        latent_shape = (1, self.encoder.out_mean.out_channels, img_shape[0],
-                        img_shape[1])
-        latent_representation = torch.randn(latent_shape).to(self.device)
+        if self.transformer is not None:
+            latent_shape = (1, self.encoder.out_mean.out_channels, img_shape[0],
+                            img_shape[1])
+        else:  # TODO: Don't hardcode shape
+            latent_shape = (1, int(self.encoder.out_mean.out_channels), 4, 4)
+        latent_representation = torch.randn(latent_shape).to(self.device).requires_grad_()
```
OleguerCanal, line 176: Not sure if I understand this if else

utilities/loader.py
```diff
@@ -59,14 +62,19 @@ def load_hgn(params, device, dtype):
...
-        **params["networks"]["transformer"],
-        dtype=dtype).to(device)
+    if "transformer" in params["networks"]:
+        transformer = TransformerNet(
+            in_channels=params["networks"]["encoder"]["out_channels"],
+            **params["networks"]["transformer"],
+            dtype=dtype).to(device)
+        decoder_in_channels = params["networks"]["transformer"]["out_channels"]
+    else:
+        transformer = None
```
OleguerCanal, line 72: Isn't that what `def instantiate_transformer` is doing?

